Issue #1, opened 2020-12-22
New Repo- DSA README file update


akrish4:
Here is the repo DSA 

Need to update the readme file @ming-tsai   @paulsonjpaul

ming-tsai:
Okis

ming-tsai:
Hi @akrish4,
Could you describe how to use and the contribution guideline?

akrish4:
i would suggest like to open a issue all the time when a new topic or program is added to a particular DSA topic so that we can avoid duplication of the codes from multiple users and maintainers can assign it to the ones who raise the issue will be given top priority to work on it.
right ? what do you say  @ming-tsai   @paulsonjpaul

ming-tsai:
I'm agreed

paulsonjpaul:
Good idea 👍

paulsonjpaul:
Add some description in the README (what this repo is about, etc.)

akrish4:
yeah , we need to update readme file

ming-tsai:
Hi @akrish4, please check the contribution guideline on readme

paulsonjpaul:
Shall we close this issue and continue the discussion in #38?

PR #11, opened 2021-01-15
docs: add contribution-guidelines

ming-tsai:
Hi @akrish4  and @paulsonjpaul,
This is related to issue #1


paulsonjpaul (changes requested):
Nice. But, where is the details about creating a new topic? For example, if I want to add a separate topic, how can I add it? Please include it as well.

ming-tsai:
> Nice. But, where is the details about creating a new topic? For example, if I want to add a separate topic, how can I add it? Please include it as well.

Check it again

PR #24, opened 2021-01-22
Add jump search java program

Amisha-Mohapatra:


Arsenic-ATG (approved):
@ming-tsai 

I would recommend going for `squash merge` on this one. like this, addition and removal of `binary-search.js`would happen under one change and will keep the commit tree relatively cleaner.

ming-tsai:
> @ming-tsai 
> 
> I would recommend going for `squash merge` on this one. like this, addition and removal of `binary-search.js`would happen under one change and will keep the commit tree relatively cleaner.

Done

PR #29, opened 2021-01-25
Add jump search py

anupam-b:


ming-tsai:
Hi @Arsenic-ATG or @akrish4,
Could you help me to check this?

Arsenic-ATG (approved):

Arsenic-ATG (approved):
functionally works
It would be better if you would have created a check too see if the input array is already sorted or not ( maybe a separate function for the same and not affect the core functionality of the `jump_search()` function ) because now it gives wrong output instead of an error messages that unsorted array was provided at first place

but as far as algorithm is concerned then it is correctly implemented and is working for all of the used cases that I have tested upon.

anupam-b:
Thank you for the feedback.
I will add a function that checks for sorted list.

Arsenic-ATG (approved):

ming-tsai:
Hi @anupam-b,
Nice work 🎉, and welcome to our team.

> you could see the invitation on this [page](https://github.com/MakeContributions)

PR #44, opened 2021-01-29
Data Structures

nightboard:
Hey there,

I added data structure directory and added singly linked list file in c++.
I also changed .gitignore to not track binary files like ( a.out and *.class for java ).
please review it and merge it.


data-structures/README.md
```diff
@@ -0,0 +1,5 @@
+# Data structures
```
Arsenic-ATG, line 1: Preferred folder name is **Linked Lists** , link we have different folder for strings, the entire repository is about data structures and algorithms.

data-structures/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+### c or c++
+
+1. [Singly Linked List](c-or-cpp/SinglyLinkedList.cpp)
```
Arsenic-ATG, line 5: After the folder name is changed to **linked lists** this can be changed to **stack implementation** (or something similar) because it is [stack](https://en.wikipedia.org/wiki/Stack_(abstract_data_type)) data structure that you are implementing here using linked list.

Arsenic-ATG (changes requested):
the program works functionally 

but have a bit of naming problems which might end up causing long term confusion once merged ( details in comments in the file)

data-structures/README.md
nightboard: Thanks for the reply i'am new so do i have to make changes and create new pull request?

paulsonjpaul: You don't need to create new pull requests. Just add more commits to the source branch in your fork (in this case, main), and the commits will be automatically added to the respective pull request.

nightboard:
Thanks for suggestions i changed as per review said please again check it if any issue.
Thanks

paulsonjpaul (changes requested):
Good job @nightboard. Just a small suggestion:

According to the [contribution guidelines](https://github.com/MakeContributions/DSA#1-folders-or-files-naming), you need to make your file and folder names lowercase and hyphenated.

Also, I see you did not change anything according to @Arsenic-ATG's second suggestion, so please check it as well.

@Arsenic-ATG, please give your feedback.

nightboard:
ok renamed directory name and @Arsenic-ATG  suggestion is to make stack implementation but i changed method names in SinglyLinkedList class so it's now convenient. if any other issues please reply me.

Arsenic-ATG (approved):

paulsonjpaul (approved):

PR #47, opened 2021-01-29
added quick sort in cpp

satcasm:


satcasm:
I am not getting it. What is the issue here can anyone help me out?

PR #48, opened 2021-01-29
added quick sort

satcasm:


sorting/c-or-cpp/quick-sort.cpp
```diff
@@ -0,0 +1,76 @@
...
+int partition(int a[],int low,int high);
+void quickSort(int a[], int low, int high);
+
+// Driver Code
+int main()
+{
+    cout<<"Enter the length of array"<<endl;
+	int n;
+	cin>>n;
+	int a[n];
```
Arsenic-ATG, line 12: Variable length array(VLA) is not a standard C++ feature ( although would still compile ) but it is highly discouraged to use it as it might end up creating rather large array on memory stack.

the possible fix for it are : -
- create a dynamic array by declaring a pointer on stack and let it point to the required allocated space on heap(using `new` operator) making the array take memory on heap instead.
- create a static array but declare it's size beforehand and make sure the user doesn't enter size of it greater than that number 
- use `std::vectors`

Arsenic-ATG (changes requested):
✅ functionally working 
⛔️ use of VLAs in the programs

sorting/c-or-cpp/quick-sort.cpp
Arsenic-ATG: same is the case with your merge sort C code also.

satcasm: My bad, I thought the main aim was just to write correct algo. I will work on it.

Arsenic-ATG: yes the main aim is on algorithm only but this will generate warnings or errors in some compilers that's why telling you about the same, I don't want someone to blame you or your code later just because it didn't ran on their system as intended just because they went overboard with the array size.

if you want, you can also get away by rid of user input at all also ( by initialising array on your own ) 
```c++
int arr[] = {12,32,3,23,12,31,32}; 
```

but using dynamic memory allocation will help you only and is also a very interesting concept to have in your knowledge inventory.👍

satcasm: Thanks for the info.

satcasm: I have modified my code feel free to give any other advices.

Arsenic-ATG (approved):

PR #86, opened 2021-02-25
Add Josephus Problem Using Circular Linked List

everly-gif:


linked-lists/c-or-cpp/josephus-problem.c
```diff
@@ -0,0 +1,63 @@
+/*Circular linked list implementation in C Programming
```
Arsenic-ATG, line 1: It would be better if you could add a bit of description of the problem on the top.

linked-lists/c-or-cpp/josephus-problem.c
```diff
@@ -0,0 +1,63 @@
...
+ printf("The Linked List is successfully created\n");
+   while(p->next!=first){
+       printf("->%d",p->data);
+       p=p->next;
+   }
+   printf("->%d",p->data);
+}
+void josephus(int k, struct Node *first){ //josephus implementation
+    struct Node *trail;
+    while(first->next!=first){
```
Arsenic-ATG, line 35: In case value of `k` is 1 
the following will lead to undefined behaviour as the value of `trail` is not initialised before hand. 
the result might end up as an infinite loop or segmentation fault.

linked-lists/c-or-cpp/josephus-problem.c
```diff
@@ -0,0 +1,63 @@
...
+        int count=1;
+        while(count!=k){
+        trail=first;
+        first=first->next;
+        count++;
+        }
+      trail->next=first->next;
+      first=trail->next;
```
Arsenic-ATG, line 43: repositioning the pointers without freeing the allocated memory they were pointing to earlier leading to potential [memory leak](https://en.wikipedia.org/wiki/Memory_leak).

fix : 
use [free()](https://en.cppreference.com/w/c/memory/free) to free the memory allocated to the executed person.

Arsenic-ATG (changes requested):
fails on a corner cases plus a memory leak.

also adding the description of the question statements (in form of comments on the top of source code) would be a fine addition.

linked-lists/c-or-cpp/josephus-problem.c
everly-gif: I have made the changes.

Arsenic-ATG (approved):

PR #88, opened 2021-02-27
Update circular.cpp

nightboard:
i improved circular linked list added efficient oop style

### DIFF
--- a/linked-lists/c-or-cpp/circular.cpp
+++ b/linked-lists/c-or-cpp/circular.cpp
@@ -1,39 +1,106 @@
 #include <iostream>
-using namespace std;
-struct Node {
-   int data;
-   struct Node *next;
+#include <unistd.h>
+
+template <typename T>
+class Node {
+    public:
+        T data;
+        Node<T>* next;
+
+        public:
+            // this create node with data
+            Node(int data)
+                : data(data) , next(nullptr) {}
 };
-struct Node* head = NULL;
-void insert(int newdata) {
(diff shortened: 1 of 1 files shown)


linked-lists/c-or-cpp/circular.cpp
```diff
@@ -1,39 +1,106 @@
 #include <iostream>
-using namespace std;
-struct Node {
-� �int data;
-� �struct Node *next;
+#include <unistd.h>
```
Arsenic-ATG, line 2: posix specific library, program will fail when run on windows. 
you might wanna remove the sleep() from your program or find an alternative of that on windows.

linked-lists/c-or-cpp/circular.cpp
```diff
@@ -1,39 +1,106 @@
...
+        }
+
+        void printList() const {
+            Node<T>* itr = this->head;
+            if(itr == nullptr) {
+                std::cout<<"List is Empty";
+            } else {
+                // print every element after 1s since it's circular
+                while(itr != nullptr) {
+                    std::cout<<itr->data<<" "<<std::flush;
```
Arsenic-ATG, line 87: avoid giving output as an infinite loop.

Arsenic-ATG (changes requested):

Arsenic-ATG:
closing this pull request as the [OP](https://github.com/nightboard) have created another PR(#89) with the proposed changes.

PR #94, opened 2021-03-02
Added Linear, Binary Search in Go

atin:


atin:
@Arsenic-ATG @ming-tsai  If you want to test the code without setting up the whole go env you can use this https://play.golang.org/

Arsenic-ATG (approved):

PR #102, opened 2021-03-06
Interval Scheduling, Quick Sort, Insertion Sort in Go

ghost:
## Interval Scheduling
For testing code - https://play.golang.org/p/JfZDZdETZc0
Also you can compare with [interval-scheduling.py](https://github.com/MakeContributions/DSA/blob/main/scheduling/python/interval-scheduling.py)

## Quick Sort
For testing code - https://play.golang.org/p/HaSDqjXlKYq
Can be compared with [quick-sort.py](https://github.com/MakeContributions/DSA/blob/main/sorting/python/quick-sort.py) and other quick sort implementations

## Insertion Sort
For testing code - https://play.golang.org/p/MP07CY8TYRo
The implementation is same as [insertion-sort.py](https://github.com/MakeContributions/DSA/blob/main/sorting/python/insertion-sort.py)

### DIFF
--- a/sorting/go/quick-sort.go
+++ b/sorting/go/quick-sort.go
@@ -1,3 +1,11 @@
+/* 
+    Quick Sort is a divide and conquer algorithm.
+    First we choose a pivot and split the array in two parts, one containing all elements less than or equal to the pivot and other contains the rest (the pivot element is in neither of them)
+    Then we recursively sort the two arrays and finally concatenate them to get the sorted array.
+    Average Time Complexity: O(n*log(n))
+    For details explanation and proof of correctness check this -> https://cs.pomona.edu/classes/cs140/pdf/l09-quicksort-proof.pdf
+*/
+
 package main
 
 import (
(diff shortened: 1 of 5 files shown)


sorting/go/quick-sort.go
```diff
@@ -0,0 +1,35 @@
+package main
```
TawfikYasser, line 1: Hi @atin I suggest here to add also a comment to explain the quick sort and its complexity.

TawfikYasser (changes requested):

sorting/go/quick-sort.go
atin: Thanks @TawfikYasser , I've added some comments explaining quick sort.

TawfikYasser: You’re welcome @atin 🙂

TawfikYasser (approved):

Arsenic-ATG (approved):

Issue #103, opened 2021-03-08
Adding circular queue


TawfikYasser:
Hi
This issue to add a `Circular Queue using Linked List` in cpp
I will open a PR
Thanks

PR #104, opened 2021-03-08
Adding the circular queue

TawfikYasser:
Hi

This PR closes issue #103 

To add the circular queue

I hope this be good

Thanks

### DIFF
--- a/queues/c-or-cpp/circular-queue-linked-list.cpp
+++ b/queues/c-or-cpp/circular-queue-linked-list.cpp
@@ -1,0 +1,107 @@
+// Content: Circular Queue using linked list
+// Author: Tawfik Yasser
+// Date: Mon, 8 Mar 2021
+#include <iostream>
+using namespace std;
+class Node
+{
+public:
+    int data;
+    Node *next;
+    Node() {}
+};
+class QueueLL
+{
+private:
+    Node *front;
+    Node *rear;
+    int _length;
+public:
+    QueueLL()
+    {
+        front = rear = NULL; _length = 0;
+    }
+    //Function to add new element in the queue
+    void enqueue(int value)
+    {
+        Node *newNode = new Node();
+        newNode->data = value;
+        newNode->next = NULL;
+        if(rear == NULL)
+        {
+            front = rear = newNode;
+            rear->next = front;
+        }
+        else
+        {
+            rear->next = newNode;
+            rear = newNode;
+            rear->next = front;
+        }
+      _length++;
+    }
+    //Function to remove element from the queue
+    void dequeue()
+    {
+        Node *temp = front;
+        if(front == NULL && rear == NULL){
+            cout<<"Empty Queue"<<endl;
+        }else if(front == rear){
+            front = rear = NULL;
+            delete temp;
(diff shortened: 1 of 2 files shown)


queues/c-or-cpp/circular-queue-linked-list.cpp
```diff
@@ -0,0 +1,107 @@
...
+            cout<<"Empty Queue"<<endl;
+        }else if(front == rear){
+            front = rear = NULL;
+            delete temp;
+        }else{
+            front = temp->next;
+            rear->next = front;
+            delete temp;
+        }
+      _length--;
```
Arsenic-ATG, line 57: in case the function is being called once the queue is already empty then it will display empty queue, but as the execution doesn't stop there, the value of `_length` will still be decreased to some negative value which should not happen

a simple fix is to return just after printing "Empty Queue"  from the function ( after line 48 )

queues/c-or-cpp/circular-queue-linked-list.cpp
```diff
@@ -0,0 +1,107 @@
...
+    int data;
+    Node *next;
+    Node() {}
+};
+class QueueLL
+{
+private:
+    Node *front;
+    Node *rear;
+    int _length;
```
Arsenic-ATG, line 18: although it's not a problem here but in future it is recommended to avoid naming identifiers starting with an underscore (`_`) as they are reserved for compiler implementation to use and using it might end up causing naming conflicts which might be difficult to diagnose as if that happens then the problem would be implantation dependent. 

it is also mentioned in C++ standards that : 
> Certain sets of names and function signatures are always reserved to the implementation:
>
> - Each name that contains a double underscore (__) or begins with an underscore followed by an uppercase letter (2.11) is reserved to the implementation for any use.
>
> - Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.

queues/c-or-cpp/circular-queue-linked-list.cpp
```diff
@@ -0,0 +1,107 @@
...
+};
+int main()
+{
+    QueueLL qll;
+    qll.enqueue(5);
+    qll.enqueue(3);
+    qll.enqueue(-2);
+    qll.print();
+    cout<<endl;
+    cout<<"Queue size = "<<qll._size();
```
Arsenic-ATG, line 99: it is recommended to add a newline character after the printing the size ( it will enhance readability as right now the next statement `qll.peek()` ends up printing stuff in same line )

Arsenic-ATG (changes requested):

Arsenic-ATG (approved):

TawfikYasser:
Can anyone merge the PR, Please?

Arsenic-ATG:
@TawfikYasser your changes have been approved and will sooner or later be merged in the main repo by @ming-tsai . 

In case you want to contribute another program and waiting for this to be merged first then, you can simply go ahead and create a separate branch to work on your new program, and feel free to create another PR regarding the same.

TawfikYasser:
Thanks @ming-tsai @Arsenic-ATG

PR #112, opened 2021-03-18
Create monotonic_binary_search.cpp

hazel11182:
This will help solve problems even if the array is not given as we use the concept of monotonic functions. Helps us acheive logn complexity instead of the brute force approach.

### DIFF
--- a/searching/c-or-cpp/monotonic_binary_search.cpp
+++ b/searching/c-or-cpp/monotonic_binary_search.cpp
@@ -1,0 +1,39 @@
+#include <iostream>
+using namespace std;
+int square_root(int a,int p){
+    int s=0;
+    int e=a;
+    int ans=-1;
+    int mid;
+    while(s<=e){
+        mid=(s+e)/2;
+        if (mid*mid==a)
+        return mid;
+        else if(mid*mid<a){
+        s=mid+1;
+        ans=mid;
+        }
+        else
+        e=mid-1;
+    }
+  float inc=0.1;
+  for(int times=1;times<=p;times++){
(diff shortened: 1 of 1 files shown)


TawfikYasser (changes requested):
I think that your code not working good

After the user enter the input `n` noting happens

Please check it again

searching/c-or-cpp/monotonic_binary_search.cpp
```diff
@@ -0,0 +1,39 @@
+#include <iostream>
+using namespace std;
+int square_root(int a,int p){
+    int s=0;
+    int e=a;
+    int ans=-1;
```
hazel11182, line 6: needs to be float ans.

TawfikYasser: Okay you can change it and re-commit

hazel11182 (commented):
two changes need to be done
1. it has to be float square_root
2. it has to be float ans=-1

TawfikYasser:
> two changes need to be done
> 
> 1. it has to be float square_root
> 2. it has to be float ans=-1

Okay no problem, You can start fixing the code

TawfikYasser:
You do not need to create new Pull Request

hazel11182:
This was my very first contribution, sorry for creating a new pull request.
what do i do now?

TawfikYasser:
> This was my very first contribution, sorry for creating a new pull request.
> what do i do now?

No problem :) , You can close this pull request and continue with the [new](https://github.com/MakeContributions/DSA/pull/114)

PR #117, opened 2021-03-23
bubble-sort.go added

UnleashMe69:
I added bubble-sort.go algorithm.
No dependencies required.

### DIFF
--- a/sorting/go/bubble-sort.go
+++ b/sorting/go/bubble-sort.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"math"
+	"os"
+	"strconv"
+	"strings"
+)
+
+//Swaps two values of a slice
+func Swap(sliceInt []int, indexValue int) {
+	//Treats edge case
+	if indexValue < len(sliceInt)-1 {
+		tempValue := sliceInt[indexValue]
+		sliceInt[indexValue] = sliceInt[indexValue+1]
+		sliceInt[indexValue+1] = tempValue
+	}
+}
+
+//Recursive implementation of Bubble sort
+func bubbleSort(sliceInt []int) {
+	for k := 0; k < len(sliceInt)-1; k++ {
+		if sliceInt[k] > sliceInt[k+1] {
+			Swap(sliceInt, k)
+			bubbleSort(sliceInt)
+		}
+	}
+}
+
+func check(e error) {
+	if e != nil {
+		fmt.Println("error: ", e)
+	}
+}
+
+func userInput() []int {
+
+	sliceInt := make([]int, 0, 10)
+	//Takes input from user
+	fmt.Println("Type in a sequence of up to 10 integers separated by a space:")
+	reader := bufio.NewReader(os.Stdin)
+	scanner, err := reader.ReadString('\n')
+	check(err)
+
+	//Trims \r and \n from the user input
+	scanner = strings.TrimSuffix(scanner, "\n")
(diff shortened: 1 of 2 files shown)


Arsenic-ATG:
@atin could you help in reviewing this pull request ? ( asking because you are the one with the most experience in golang among us )

sorting/go/bubble-sort.go
```diff
@@ -0,0 +1,73 @@
...
+		sliceInt[indexValue] = sliceInt[indexValue+1]
+		sliceInt[indexValue+1] = tempValue
+	}
+}
+
+//Recursive implementation of Bubble sort
+func bubbleSort(sliceInt []int) {
+	for k := 0; k < len(sliceInt)-1; k++ {
+		if sliceInt[k] > sliceInt[k+1] {
+			Swap(sliceInt, k)
```
atin, line 26: Why make a different function for swap? You can simple do it it one line like this `a, b = b, a`
Also the `if` condition in the swap function is always true because `k < len(sliceInt)-1` in the for loop
So I think you should remove `Swap` function and replace `Swap(sliceInt, k)` with `sliceInt[k], sliceInt[k+1] = sliceInt[k+1], sliceInt[k]`

sorting/go/bubble-sort.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func check(e error) {
+	if e != nil {
+		fmt.Println("error: ", e)
+	}
+}
+
+func userInput() []int {
```
atin, line 38: Also the `userInput()` part is making the code too complex, the aim of this code should be demonstrating how bubble sort works so I suggest you should completely remove this part and take a hardcoded slice and perform bubble sort on that.

sorting/go/bubble-sort.go
```diff
@@ -0,0 +1,73 @@
+package main
```
atin, line 1: You should also add a little info about the algo itself like time complexity and a little explanation.

atin (changes requested):

atin:
@UnleashMe69 here is a reduced version of your code, https://play.golang.org/p/2nQL9nF0bLf
@Arsenic-ATG you can also run and check if it works correctly.

Arsenic-ATG:
> @UnleashMe69 here is a reduced version of your code, https://play.golang.org/p/2nQL9nF0bLf
> @Arsenic-ATG you can also run and check if it works correctly.

looks good to me, also works as intended ✅

UnleashMe69:
Based on your reviews, here is the new version based on your suggestions: 
[https://play.golang.org/p/u2rOrKFCNhT](https://play.golang.org/p/u2rOrKFCNhT)

atin:
> Based on your reviews, here is the new version based on your suggestions:
> https://play.golang.org/p/u2rOrKFCNhT

Looks good! @ming-tsai @Arsenic-ATG can you also check and merge it?

ming-tsai:
Hi @atin,
Could you check again, he had pushed a new commit

atin (approved):

atin:
> Hi @atin,
> Could you check again, he had pushed a new commit

@ming-tsai looks good!

ming-tsai (approved):

ming-tsai:
Nice work @atin,and @UnleashMe69

ming-tsai:
Thanks for your contributions @UnleashMe69
> The invitation [link](https://github.com/MakeContributions)

PR #122, opened 2021-03-28
Added some sorting algorithms using arrays and added reverse a stack …

KishkinJ10:
…algorithm using stack

### DIFF
--- a/arrays/c-or-cpp/BinarySearch.cpp
+++ b/arrays/c-or-cpp/BinarySearch.cpp
@@ -1,0 +1,38 @@
+//Algorithm Binary Search using Arrays
+
+#include<iostream>
+using namespace std;
+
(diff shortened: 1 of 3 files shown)


Arsenic-ATG (changes requested):
linear search and binary search has already been implemented in this repository in C++. 

as for stack reversing program is considered, so in order for that to qualify, you need to update the [index readme](https://github.com/KishkinJ10/DSA/blob/main/stacks/README.md) with your creation in the list.

Arsenic-ATG:
I am flagging this post as inactive, if the OP doesn't respond within 24 hours, this PR would be closed due to inactivity.

Arsenic-ATG:
closing this PR for long inactivity.

Issue #139, opened 2021-04-05
Wanted to add Prim's algorithum in C Language


UG-SEP:
Hey,
I saw that there is  absence of the one of the famous Prim's algorithm in Graph->C-or-CPP so can i add this algorithm basically Prim's algorithm is a way to implement the Minimal Spanning Tree (MST) `means to find the minimum cost way to cover all the vertex`.
Please Assign Me

PR #140, opened 2021-04-07
Prim's Algorithm in c

UG-SEP:
## Description
Hey I have added prim's algorithm in c language.
Please review it @Arsenic-ATG and @TawfikYasser 
## Related Issue 
 fixes: #139

### DIFF
--- a/graphs/c-or-cpp/Prim's-algorithm.c
+++ b/graphs/c-or-cpp/Prim's-algorithm.c
@@ -1,4 +1,5 @@
-// This is Prim's algorithm to find the lowest cost to traverse it all vertex of a graph or MST(Minimal Spanning Tree)
+/*                                                               Prim's Algorithm                                               
+             This is Prim's algorithm to find the lowest cost to traverse it all vertex of a graph or MST(Minimal Spanning Tree)                                              */
 #include<stdio.h>
 #include<stdlib.h>
 // node type structure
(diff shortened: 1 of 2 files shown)


graphs/c-or-cpp/Prim's-algorithm.c
```diff
@@ -212,4 +214,6 @@ enter the node no of v2 and weight 3
 2
 Output: Cost: 4
 
+Time Complexity:O(ElogV)
```
UG-SEP, line 217: time complexity is write or wrong

TawfikYasser: Depends on the way of **implementation** for the algorithm

For Binary heap yes it is `O(ElogV)`
For Array it is `O(V^2)`
For Fibonacci heap it is `O(E + VlogV)`

meenagupta5: > Depends on the way of **implementation** for the algorithm
> 
> For Binary heap yes it is `O(ElogV)`
> For Array it is `O(V^2)`
> For Fibonacci heap it is `O(E + VlogV)`

Using priority queue

Arsenic-ATG: Time complexity will be `O(V^2)` in this particular implementation. Because it is just using linked list to store the values and traversing the entire linked list to find the minimum value in the set of edges ( which is similar to using arrays ) 

@meenagupta5 all the options @TawfikYasser pointed out will use priority queue only, the difference lies in how user is implementing that priority queue. ( a typical implementation uses min-heap for prim's algo )

Arsenic-ATG (approved):

ming-tsai:
Nice work @UG-SEP

UG-SEP:
> Nice work @UG-SEP

thanks a lot

PR #143, opened 2021-04-08
Added program in c for printing unique elements in an array

Dabohemianstarlord:


arrays/c-or-cpp/unique-elements-in-an-array.c
```diff
@@ -0,0 +1,32 @@
+#include<stdio.h>
+
+int main(){
+    int a[7];
```
Arsenic-ATG, line 4: What if an array of bigger size is to be evaluated ?

arrays/c-or-cpp/unique-elements-in-an-array.c
```diff
@@ -0,0 +1,32 @@
...
+    
+    for(int k=0; k<7; k++){
+        scanf("%d", &a[k]);
```
Arsenic-ATG, line 7: prefer to either ask user for size before, or create a default array with the some sample values to test this algo on.

arrays/c-or-cpp/unique-elements-in-an-array.c
```diff
@@ -0,0 +1,32 @@
...
+    int flag;
+
+    for(int i=0; i<7; i++){
+        flag = 0;
+        
+        for(int j=0; j<=7; j++){
+
+            if(i != j){
+            
+                if(a[i] == a[j]){
```
Arsenic-ATG, line 18: these 2 if statements can be combined in one to make the code a bit shorter.

```c++
if ( ( i != j ) && ( a[i] == a[j] ) )
    flag++;
```

Arsenic-ATG (changes requested):
overall working but there are some minor problems ( consider them as nitpicks ) which are mentioned in review comments.

ming-tsai:
Hi @Dabohemianstarlord,
Could you try to fix the comments?

UG-SEP:
@ming-tsai  can i review it

ming-tsai:
> @ming-tsai  can i review it
> 

Sorry @Arsenic-ATG is reviewing

UG-SEP:
> > @ming-tsai  can i review it
> 
> Sorry @Arsenic-ATG is reviewing

ok no problem 
next time

arrays/c-or-cpp/unique-elements-in-an-array.c
cclauss: Does this code properly deal with negative numbers?

Arsenic-ATG: @cclauss yes it does.

arrays/c-or-cpp/unique-elements-in-an-array.c
```diff
@@ -0,0 +1,34 @@
+#include<stdio.h>
+
+int main(){
+    int a[1000];
```
Arsenic-ATG, line 4: Prefer creating it a dynamic array and create it according to what size user enters.

arrays/c-or-cpp/unique-elements-in-an-array.c
```diff
@@ -0,0 +1,34 @@
...
+    printf("Enter size of the array: ");
+    scanf("%d", &size);
+    
+    printf("Enter elements of the array: ");
+    for(int k=0; k<size; k++){
+        scanf("%d", &a[k]);
+    }
+    int flag;
+
+    for(int i=0; i<7; i++){
```
Arsenic-ATG, line 15: You changed the input to work with array of size upto 1000 but your loops will only perform 7 iterations each no matter the size of the array.

Arsenic-ATG (changes requested):

arrays/c-or-cpp/unique-elements-in-an-array.c
cclauss: `"Enter size of the array: " -5`

UG-SEP: > @cclauss yes it does.

@Arsenic-ATG according to the code written it does take negative number as size if he wanted to do that he can check that the given size is negative if yes size=size*-1;

Arsenic-ATG: @cclauss oh, I though you were talking about negative numbers as array elements.

Well the main focus of this program( and I think all the other programs here ) is implementation of algorithm ( finding unique elements in an array ), considering the inputs are given correctly and in reasonable range, not able to handle negative size is actually not the algorithm's fault. I don't expect people who are visiting this repo to use these programs as they are, they are mainly there for educational and learning purposes, so that anyone can view the program and learn about how the algorithm is being implemented in a particular language, and adding additional checks to check for bounds of algo can make it overwhelming for newcomers especially when the algorithm is implemented directly in main() and not as a different function.

But yes, I do agree that it's a good practice to keep it free from all human errors also and adding those additional checks will always be appreciated.

Arsenic-ATG (approved):

arrays/c-or-cpp/unique-elements-in-an-array.c
cclauss: I know that this is already merged but please allow me to make a few comments...

> when the algorithm is implemented directly in main()

I believe that is an anti-pattern because such an algorithm is very difficult to move into a larger program and is complicated to test.  I think my contributions over the past week have demonstrated the value of testing algorithms.  Without tests, even simple algorithms deliver results that are different from our intuition.  (See the comments on my PRs.). Algorithms that come with tests are more dependable and can be modified/improved with confidence.  Writing tests help us to find problems before our users do.

> considering the inputs are given correctly and in reasonable range

This is called [“happy path”](https://en.wikipedia.org/wiki/Happy_path) programming and it is a really poor pattern to teach to newcomers.  We should encourage contributors (even first-time contributors) to be on the lookout for corner cases that will make software fragile in the real world.  We do not our contributors to fail future job interviews because they only know how to code for the happy path.

ming-tsai: Good point

Arsenic-ATG: Ok @cclauss, I will be stricter while reviewing the pull requests from now on 🙂👍

and as for this pull request is concerned, I will be updating the program soon to meet the requirements.

PR #159, opened 2021-04-11
Add Python doctests to count-inversions.py

cclauss:
These [doctests](https://docs.python.org/3/library/doctest.html) can be run with `python3 -m doctest -v count-inversions.py`

Our GitHub Actions will also run these doctests.  https://github.com/MakeContributions/DSA/actions

The answer to the last test should be 0, not 2.  The algorithm is currently counting inversions that are not required.

@atin Your review please.

### DIFF
--- a/arrays/python/count-inversions.py
+++ b/arrays/python/count-inversions.py
@@ -2,6 +2,8 @@
 Algorithm Type: Divide & Conquer
 Time Complexity: O(n*log(n))
 """
+numbers = [8, 2, 1, 5, 7, 3, 9, 2, 0, 1]
+
 
 def count_split_inv(arr, left, right):
   split_inv = ridx = lidx = 0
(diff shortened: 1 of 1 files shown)


aayushjain7 (dismissed):

atin:
> The answer to the last test should be 0, not 2. The algorithm is currently counting inversions that are not required.

Yes, there was a small mistake in line 15, instead of `right[ridx] <= left[lidx]` it should be `right[ridx] < left[lidx]`, @cclauss can you change this in this PR only?

arrays/python/count-inversions.py
```diff
@@ -23,9 +25,22 @@ def count_split_inv(arr, left, right):
...
 
-def count_inversions(arr):
+def count_inversions(arr) -> int:
+  """
+  Sort arr and return the number of inversions required.
+
+  >>> numbers
+  [8, 2, 1, 5, 7, 3, 9, 2, 0, 1]
+  >>> count_inversions(numbers)
+  30
```
atin, line 38: This should be `28` the problem in line 15 also affects this

atin (changes requested):

arrays/python/count-inversions.py
cclauss: Perfect!

aayushjain7 (approved):

atin (approved):

aayushjain7 (approved):

ming-tsai (approved):

PR #161, opened 2021-04-11
Add Python doctests to karatsuba.py

cclauss:
`subtract()` can give the wrong answer when `y` > `x`.  @atin Your review, please.

Also, [`sum()`](https://docs.python.org/3/library/functions.html) is a Python built-in with a different calling convention so it would be better to use the function name `add()`.

### DIFF
--- a/multiplication/python/karatsuba.py
+++ b/multiplication/python/karatsuba.py
@@ -49,7 +49,8 @@ def add(x, y):
 def absolute_difference(x, y):
   """
   >>> all(absolute_difference(str(x), str(y)) == str(abs(x - y)) for x, y
-  ...   in ((0, 0), (0, 1), (1, 1234567890), (543210, 9876543)))
+  ...   in ((0, 0), (0, 1), (1, 0), (1, 1234567890), (543210, 9876543)))
+  True
   >>> absolute_difference("1", "0")
   '1'
   """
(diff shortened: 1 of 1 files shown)


aayushjain7 (approved):

aayushjain7 (dismissed):

aayushjain7:
LGTM!

atin:
> `subtract()` can give the wrong answer when `y` > `x`. @atin Your review, please.

Actually it subtracts the smaller number from larger number, hence if `y` > `x` it'll do `y-x` not `x-y`. Though modifying it to always do `x-y` won't affect the code because it's only used at one place in the code and  `x` is always greater than `y` in that case. So I suggest either we change the function name to something more meaningful or change the implementation do always do `x - y`. Both are fine by me. For the 2nd option we can just do `return result if x > y else '-' + result` I think.

> Also, [`sum()`](https://docs.python.org/3/library/functions.html) is a Python built-in with a different calling convention so it would be better to use the function name `add()`.

Yes, you're right. Thanks for pointing it out.

cclauss:
Perhaps `subtract()` could be renamed to `absolute_difference()` and the test could be `str(abs(x - y))`

atin:
Yeah, that will be good.

aayushjain7 (approved):

ming-tsai (approved):

PR #162, opened 2021-04-11
Add Python doctest to interval-scheduling.py

cclauss:
I do not understand the output of this algorithm.  Is there a URL that explains its results?

`[5, 4, 1, 7]` shows that the optimal schedule is four batches but which jobs should be run in each batch?

A [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) would improve the readability of the code...
```
from collections import namedtuple

Job = namedtuple("Job", "job_id start_time finish_time")
```
@atin Your review, please.

### DIFF
--- a/scheduling/python/interval-scheduling.py
+++ b/scheduling/python/interval-scheduling.py
@@ -5,29 +5,36 @@ The Strategy: At each step choose the job with earliest finish time
 Algorithm Type: Greedy
 Time Complexity: O(n*log(n))
 """
+jobs = [
+  (0, 2, 8),  # (job_id, start_time, finish_time)
+  (1, 6, 10),
+  (2, 1, 3),
+  (3, 4, 7),
+  (4, 3, 6),
+  (5, 1, 2),
+  (6, 8, 10),
+  (7, 10, 15),
+  (8, 12, 16),
+  (9, 14, 16)
+]
 
-def get_opt_schedule(jobs): # Returns the job_id's in the optimal schedule
+
+def get_opt_schedule(jobs):
+  """
+  Return the job_id's in the optimal job schedule
+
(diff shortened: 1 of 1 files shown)


aayushjain7 (approved):

atin:
> I do not understand the output of this algorithm. Is there a URL that explains its results?

This explains the algorithm.
https://en.wikipedia.org/wiki/Interval_scheduling#Greedy_polynomial_solution
> `[5, 4, 1, 7]` shows that the optimal schedule is four batches but which jobs should be run in each batch?

Not sure what you mean by this, the function returns the job ids of the optimal jobs.
> A [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) would improve the readability of the code...

Ok, thanks. Didn't knew about namedtuple before.

scheduling/python/interval-scheduling.py
```diff
@@ -5,29 +5,37 @@
...
+]
+
 
-def get_opt_schedule(jobs): # Returns the job_id's in the optimal schedule
+def get_opt_schedule(jobs):
+  """
+  Return the job_id's in the optimal jobs
+  https://en.wikipedia.org/wiki/Interval_scheduling#Greedy_polynomial_solution
+
+  >>> get_opt_schedule(jobs)  # Result shows four batches but which jobs in each batch?
```
atin, line 27: > Result shows four batches but which jobs in each batch?

What do you mean by this? The algorithm finds the max possible number of non overlapping jobs and returns their job id's.

atin (approved):

aayushjain7 (approved):

ming-tsai (approved):

Issue #141, opened 2021-04-07
Program to find total no. of words in sentence 


UG-SEP:
Hey,
I saw that there is no program related to Problem:
Total Number of Words in [string->c-or-cpp](https://github.com/MakeContributions/DSA/tree/main/strings/c-or-cpp)
`Input:`  The DSA is the most important topic in programming to solve real world problems
`Output:`  There are 14 words in the sentence
Please assign me the issue 
Thanks

PR #185, opened 2021-04-14
Total number of words in a sentence

UG-SEP:
Hey,
I have added total number of word in sentence program in c language
Please review it @Arsenic-ATG 
Closes: #141 
Thanks


UG-SEP:
@ming-tsai  please change the label from c++ to c

cclauss:
Nice!  A cool second algorithm for a future pull request would be the number of ___unique___ words in a sentence.

strings/c-or-cpp/count-words.c
```diff
@@ -0,0 +1,66 @@
...
+#include<stdio.h>
+#include<string.h>
+#include<stdlib.h>
+
+//Function which calculate and return total numbers of words
+int count_words(char *sen)
+{
+    int i=0,check=1,word=0;
+    //dynamically allocating memory
+    realloc(sen,strlen(sen)+1);
```
Arsenic-ATG, line 11: Return value of realloc is being ignored here. 

also there is no need for reallocation here as you have already done it one before( line 49 ).

strings/c-or-cpp/count-words.c
```diff
@@ -0,0 +1,66 @@
...
+//driver code
+int main()
+{
+    //initialize required variables
+    char *sen,res;
+    // dynamically locating the address in sen variable
+    sen=(char*)malloc(sizeof(char)*1000);
+    printf("Enter a sentence: ");
+    //taking input
+    fgets(sen,1000,stdin);
```
Arsenic-ATG, line 47: line 44
```c
sen=(char*)malloc(sizeof(char)*1000);
```

this may return NULL in case malloc fails to allocate memory, in that case trying to access a null pointer via `fgets()` will lead to undefined behaviour.

fix is to perform a quick check on the return value of `malloc()` and exit the program with an error message in case it returns NULL.

something like this 
```c
if(sen == NULL)
{
    fprintf(stderr, "malloc() failed to allocate memory\n");
    return 1;
}
```

Arsenic-ATG (changes requested):

strings/c-or-cpp/count-words.c
UG-SEP: yah sorry didn't see it thanks for telling me

strings/c-or-cpp/count-words.c
UG-SEP: yes you are right

Arsenic-ATG (approved):

PR #192, opened 2021-04-14
Added functionality to merge two singly linked list

LightMonarch:


aayushjain7 (changes requested):
LGTM! But remove extra white spaces in between the lines.
Also, update the README.md file of the corresponding topic.

LightMonarch:
Alright. Done

I have some doubts about my code i don't understand,. Whenever i try to `delete oldhead;` ( line 45 & 39) it some how changes the `l1` list after the merge (line 98) even though i am passing it by const reference and copying it in the next line. When i don't
delete oldhead it works the way it should be.

why is that?

linked-lists/c-or-cpp/merge.cpp
```diff
@@ -0,0 +1,97 @@
...
+		void popfront();
+		void display();
+		void merge(const LinkedList<T>& list2) ;
+	private:
+		struct node{
+			T data;
+			node* next;
+			node(const T& data): data(data), next(nullptr){}
+		};
+		node* _head = nullptr;
```
Arsenic-ATG, line 18: Avoid starting the name of any identifier with underscore ("_"), as they are reserved for implementation to use. Using it might cause conflict in case your compiler is also using the same name.

linked-lists/c-or-cpp/merge.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+template <typename T>
+void LinkedList<T>::popfront(){
+	//sets the head pointer to nullptr and the next node as the new head
+	
+	//if there is no node then do nothing
+	if(_head == nullptr) return;
+	if(_head->next == nullptr){
+		_head = nullptr;
+		#delete _head;
```
Arsenic-ATG, line 36: invalid preprocessor directive `#delete`
same on line 42.

also if this is `delete` operator then it should be before assigning head to `nullptr`, as now it is just deleting a null pointer and previous statement is causing a memory leak.

linked-lists/c-or-cpp/merge.cpp
```diff
@@ -0,0 +1,97 @@
...
+	node* thru = _head;
+	while(thru != nullptr){
+		std::cout << thru->data << std::endl;
+		thru = thru->next;
+	}
+}
+
+template <typename T>
+void LinkedList<T>::merge(const LinkedList<T>& other) {
+	LinkedList<T> l1 = other;
```
Arsenic-ATG, line 72: As there is no user-defined assignment operator in this class so default operator of the class will operate on here, which will just copy all the contents of `other` to `l1` ( **including the location where head pointer is pointing to** ), this will make the changes that you made to `l1` cause undesired changes in `other` list too.

---

You can fix this problem in 2 ways :- 

1. Create your own custom overloaded assignment operator(`=`) for the class `LinkedList` which does a deep copy of the list.

2. Perform the copy manually here.

Arsenic-ATG (changes requested):
> I have some doubts about my code i don't understand,. Whenever i try to delete oldhead; ( line 45 & 39) it some how changes the l1 list after the merge (line 98) even though i am passing it by const reference and copying it in the next line. When i don't
delete oldhead it works the way it should be.

the reason for this is in line 72 of your program. ( check my review comment for more info )

aayushjain7 (changes requested):
Make necessary changes as suggested by @Arsenic-ATG

linked-lists/c-or-cpp/merge.cpp
LightMonarch: Okay i replaced it with "head_" for private naming convention

linked-lists/c-or-cpp/merge.cpp
LightMonarch: got it, #delete is a mistake on my part, i confused it with python commenting.

linked-lists/c-or-cpp/merge.cpp
LightMonarch: That bug made me go crazy.
Thanks for the insights.

LightMonarch:
@aayushjain7  and @Arsenic-ATG  everything works perfectly now.
Please check...

aayushjain7 (dismissed):
LGTM.

Arsenic-ATG (approved):

ming-tsai:
Hi @LightMonarch,
Nice work, and welcome to our team
> The invitation [link](https://github.com/MakeContributions)

LightMonarch:
Thank you. Good to be part of the community

PR #194, opened 2021-04-14
Python: Total number of words in a sentence

cclauss:


sridhar-5 (changes requested):
code looks good to me but please change the file name. `_` should be replaced with `-`.

cclauss:
I do not think that we should be encouraging our contributors to create files that cannot be imported.  [Hyphens break imports in Python](https://stackoverflow.com/questions/2740026/why-are-underscores-better-than-hyphens-for-file-names) because valid Python identifiers cannot contain hyphens.  Now that we are moving to programming language-based root directories, my sense is that each directory should track the filename conventions of that language.  Thoughts?

aayushjain7 (approved):

sridhar-5:
In that case this can be approved. I just followed the PR documentation of this repository.

sridhar-5 (approved):

PR #202, opened 2021-04-15
Add sorted insert method in linked-list in C++

LightMonarch:


linked-lists/README.md
```diff
@@ -1,27 +0,0 @@
-# Linked Lists
```
Arsenic-ATG, line 1: looks like you accidentally deleted the entire index readme for linked list.

linked-lists/c-or-cpp/sortedInsert.cpp
```diff
@@ -0,0 +1,111 @@
...
+void LinkedList<T>::pushfront(const T& data){
+	node* n = new node(data);
+	n->next = head_;
+	head_ = n;
+}
+
+template <typename T>
+void LinkedList<T>::push(const T& data){
+	if(!sortInsertInit_){
+		std::cout << "Initialize sortInsert before calling push" << std::endl;
```
Arsenic-ATG, line 54: If triggering sortedInsert() in a mandatory for push() to work then what is the reason for it to be an additional function ? 

you could have passed it as a boolean value as an argument to the push() function only. Also, if boolean value represents the list being ascending or descending then condition `if(!sortInsertInit_)` will always be true for descending list also.

---

Prefer using stderr ( instead of stdout ) to display error messages. or throw exceptions in such cases.

linked-lists/c-or-cpp/sortedInsert.cpp
```diff
@@ -0,0 +1,111 @@
+#include <iostream>
+#include <time.h>
```
Arsenic-ATG, line 2: Avoid using C-style header in C++, change it `<ctime>`.

Arsenic-ATG (changes requested):
apart from the changes pointed out in review comments, consider [cleaning your commits](https://christoph-rumpel.com/2015/05/clean-up-your-commits-for-a-pull-request) for this pull request. There are a lot of unnecessary commits here.

linked-lists/c-or-cpp/sortedInsert.cpp
LightMonarch: would it be better if i do something like this?
`l1.sortInsertRev = true`
 this is for descending order, so by default it will be in ascending order.

So if I want to it to be that way it has to be set to `true` after creating the list before pushing anything.
If i do it after i push some integers the list order will be messed up, some integers will be in ascending and some will be in descending.
Should i avoid such events and write an if statement or leave it like that?

Issue #201, opened 2021-04-15
Good or Bad String in C++


UG-SEP:
Hey,
 I wanted to add a program which let you know whether the string is good or bad 
Criteria:
If the string contain continuously more than  5 vowel or continuously more than 3 constant then  it is a bad string else it is  a good string 
Question of GeeksforGeeks Platform: [Bad or Good String](https://practice.geeksforgeeks.org/problems/good-or-bad-string1417/1/?company[]=Amazon&company[]=Amazon&problemStatus=solved&problemType=functional&page=1&sortBy=submissions&query=company[]AmazonproblemStatussolvedproblemTypefunctionalpage1sortBysubmissionscompany[]Amazon)

Arsenic-ATG:
hey @UG-SEP, also mention the programming language you would be using to attempt this algorithm. 

It will make this issue open for others also to contribute but in different languages.

PR #210, opened 2021-04-15
Create is_good_str.py

cclauss:
As discussed in #201


aayushjain7 (changes requested):
Update README.md file.

aayushjain7 (approved):
looks good!

PR #252, opened 2021-04-22
 Allocate minimum number of pages

aayushjain7:


algorithms/Java/searching/allocate-min-pages.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.io.*;
+import java.util.*;
+class GFG {
+	public static void main (String[] args) {
+		Scanner sc=new Scanner(System.in);
+		int t=sc.nextInt();  //number of test cases
+		while(t-->0){
+		    int N=sc.nextInt();
+		    int pages[]=new int[n];
```
TawfikYasser, line 16: Hi @aayushjain7 , Please check what is `n` in the following: ```java new int[n] ```

TawfikYasser (changes requested):

algorithms/Java/searching/allocate-min-pages.java
aayushjain7: Oh. I'll correct it.

TawfikYasser (approved):
Nice work @aayushjain7

Issue #277, opened 2021-04-29
Wanted to add Ternary Search in c and c++


UG-SEP:
Hey @ming-tsai ,
I wanted to [ternary search](https://en.wikipedia.org/wiki/Ternary_search)
Please assign me to do that 
Thanks

PR #280, opened 2021-05-03
chore(C & CPP): Ternary Search

UG-SEP:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Added Ternary Search in c and c++ language and also updated the readme of c and CPP

**Other information:**
closes: #277

### DIFF
--- a/algorithms/CPlusPlus/Searching/Ternary-search.cpp
+++ b/algorithms/CPlusPlus/Searching/Ternary-search.cpp
@@ -45,8 +45,11 @@ int main()
     cout<<"Enter size: ";
     // taking input
     cin>>n;
+    // if size is negative then make it positive one
+    if(n<0)
+       n=n*-1;
     // dynamically allocating array
-    arr=new int[n];
+    arr=(int*)malloc(sizeof(int)*n);
     for(i=0;i<n;i++)
     {
         cin>>arr[i];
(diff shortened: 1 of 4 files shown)


algorithms/CPlusPlus/Searching/Ternary-search.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+// driver code
+int main()
+{
+    int n,*arr,i,to_search,res;
+    cout<<"Enter size: ";
+    // taking input
+    cin>>n;
+    // dynamically allocating array
+    arr=new int[n];
```
Arsenic-ATG, line 49: Just a piece of advice : If you are opting for C++, then it is always a better option to go for stl containers and avoid C idioms as much as possible.

I see no reason to not use C-style arrays over C++ standard vectors in this case.

Arsenic-ATG (changes requested):
functionally works but need some minor changes before being merged.( see review comments for details )

algorithms/CPlusPlus/Searching/Ternary-search.cpp
UG-SEP: ok I will remember this but I haven't learned STL right now

Arsenic-ATG (approved):

PR #297, opened 2021-05-13
chore(CPlusPlus): add binary search tree

Ashborn-SM:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Binary search tree algorithm in c++ and update that in README.md

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Trees/binary-search-tree.cpp
+++ b/algorithms/CPlusPlus/Trees/binary-search-tree.cpp
@@ -1,0 +1,114 @@
+#include <iostream>
+
+//Structure of the Tree
+struct TreeNode{
+	int data;
+	TreeNode* left;
+	TreeNode* right;
+	TreeNode(const int& data): data(data), left(nullptr), right(nullptr){}
+};
+
+TreeNode* find(TreeNode* root, const int& data){
+	/**
+	 * Find the node that contains the given data and
+	 * return that node
+	 *
+	 * @params: `root` root/parent node of the tree
+	 * @params: `data` data to be find in the tree
+	 * @return: tree node that contains the data
+	 *
+	 * Average case Time Complexity: O(log(n))
+	 * Worst case Time Complexity: O(n)
+	 *
+	 */
+
+	if(root == nullptr) 	    { throw std::runtime_error("Error: find() cannot find the data. The data doesn't exist."); }
+	else if(root->data == data) { return root; }
+	else if(root->data < data)  { return find(root->right, data); }
+	else 			    { return find(root->left, data); }
+}
+
+void Insert(TreeNode*& root, const int& data){
+	/**
+	 * Create and Insert the node in the appropriate place of the tree
+	 *
+	 * @params: `root` root/parent node of the tree
+	 * @params: `data` data to be inserted in the tree
+	 * @return: void
+	 *
+	 * Average case Time Complexity: O(log(n))
+	 * Worst case Time Complexity: O(n)
+	 *
+	 */
+
+	if(root == nullptr) 	    { root = new TreeNode(data); }
+	else if(root->data == data) { throw std::runtime_error("The node already exist. Duplicates not allowed"); }
+	else if(root->data < data)  { Insert(root->right, data); }
+	else 			    { Insert(root->left, data); }
+}
+
+void print(TreeNode* root){
+	/**
+	 * Print the tree in an inorder fashion
+	 * 
+	 * @params: `root` root/parent node of the tree
+	 * @return: void
+	 */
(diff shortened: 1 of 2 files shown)


ming-tsai:
Hi @Ashborn-SM,
This is not documentation, the documentation check should mark when you change some file on [docs](https://github.com/MakeContributions/DSA/tree/main/docs) folder

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -0,0 +1,96 @@
...
+	   /\	/ 	
+	  2 11  20
+
+	  Output: 2 4 11 19 20 22 37 42 51 55
+
+	*/
+
+	TreeNode* n = find(root, 2);
+	std::cout << "\nValue of n: " << n->data << std::endl; //Output: Value of n: 2
+	return 0;
```
Arsenic-ATG, line 95: Your `root` will leak here. 

In the current scenario it would be ok, but will become prominent as soon as this implementation of data structure is used somewhere else.

my recommendation would be to either perform a manual cleanup inside the destructor of the `TreeNode` structure, or make use of smart pointers and let them manage memory for you.

Arsenic-ATG (changes requested):

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
Ashborn-SM: what do you mean leak? 
how does it affect the code?
and what exactly should i do in the destructor to avoid it ?

could you please elaborate for me

Arsenic-ATG: @Ashborn-SM  let's say we have a function 

```c++
void func()
{
    TreeNode* treeptr = nullptr;
    Insert(root, 1);
    Insert(root, 2);
    Insert(root, 3);
    Insert(root, 4);
    .
    .
    .
    .
    // and so on
}
```

and now as soon as `treeptr` goes out of scope, it would be deleted from the stack memory leading to nothing pointing to the data you stored in heap that you allocated dynamically while inserting the nodes to the tree which is now inaccessible to your program anymore and thus never be released till the time program is running, this is a classical example of a [memory leak](https://en.wikipedia.org/wiki/Memory_leak). 

One fix is to make sure to delete the memory you allocated before the pointer pointing to the root gets destroyed, and the best place to do the cleanup is obviously the destructor.

alternatively, you can also use C++'s smart pointers by creating every node a [unique pointer](https://en.cppreference.com/w/cpp/memory/unique_ptr) and use [`std::make_unique()`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) for construction of a node. It will not only handle node's dynamic lifetime for you but will also ensure exception safety of your structure which you have to manage manually the other way as destructor is never called if the thrown exception is never caught anywhere.

Ashborn-SM: i know how to manually de-allocate memory but i don't know how do it in destructor.
To de-allocate i will be needing the root of the tree, how can i get that in the destructor?

Arsenic-ATG: an optimal way of doing this would be to encapsulate the entire functionality( insert(), find() and print() )  and data ( node structure ) in a class and perform destruction of the tree object in the destructor of that class. 

but for now you can just create a function( taking root as its param ) that will delete the entire tree and call it just before the end of main() and maybe work on the above plan in another pull request as an enhancement some other time.

Ashborn-SM: alright.
is there any way to check that i successfully de-allocated the memory in the heap

Arsenic-ATG: nope, there is no standard way (as of now) to check if all the memory is deallocate or not, generally ppl make the pointer they deallocated point to nullptr to know that it is pointing to nothing and can perform the checks later.

That's why C/C++ programs are highly vulnerable to memory leaks as they don't perform automatic garbage collection and put everything in the hands of dev. That's the only reason they introduces smart pointers after C++11.

Arsenic-ATG (approved):

ming-tsai:
Nice work @Ashborn-SM and @Arsenic-ATG

Ashborn-SM:
thanks @ming-tsai and @Arsenic-ATG  for helping me.
i will try my best to contribute more

PR #298, opened 2021-05-13
chore(Python): create heap-sort

Sranu2109:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

Created heap sort in python.

### DIFF
--- a/algorithms/Python/sorting/heap-sort.py
+++ b/algorithms/Python/sorting/heap-sort.py
@@ -1,6 +1,7 @@
 # Heap sort in python
 from typing import Callable
 
+
 def heap_data(nums, index, heap_size):
     largest_num = index
     left_index = 2 * index + 1
--- a/algorithms/Python/sorting/heap-sort.py
+++ b/algorithms/Python/sorting/heap-sort.py
@@ -1,6 +1,7 @@
 # Heap sort in python
 from typing import Callable
 
+test_arr = [10, 1, 6, 256, 2, 53, 235, 53, 1, 7, 0, -23, 23]
 
 def heap_data(nums, index, heap_size):
     largest_num = index
@@ -26,13 +27,9 @@ def heap_sort(nums):
     return nums
 
 
-user_input = input("Enter the list of numbers separated by a comma:\n").strip()
-nums = [int(item) for item in user_input.split(',')]
-print('Sorted list is: ')
-heap_sort(nums)
-print(nums)
-print('Time complexity : Best case = Avg case = Worst case = O(n logn)')
-
+if __name__ == "__main__":
+    print("Sorted Array:", heap_sort(test_arr))
+    
 
 # Runtime Test Cases:-
 # Test case 1.
--- a/algorithms/Python/sorting/heap-sort.py
+++ b/algorithms/Python/sorting/heap-sort.py
@@ -5,23 +5,23 @@ def heapsort(alist):
     for i in range(len(alist) - 1, 0, -1):
         alist[0], alist[i] = alist[i], alist[0]
         max_heapify(alist, index=0, size=i)
- 
+
 def parent(i):
     return (i - 1)//2
- 
+
 def left(i):
     return 2*i + 1
- 
+
 def right(i):
     return 2*i + 2
- 
+
 def build_max_heap(alist):
     length = len(alist)
     start = parent(length - 1)
     while start >= 0:
         max_heapify(alist, index=start, size=length)
         start = start - 1
- 
+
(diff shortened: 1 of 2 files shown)


sridhar-5:
I'm checking it now.

algorithms/Python/sorting/heap-sort.py
```diff
@@ -0,0 +1,57 @@
...
+        largest = index
+    if (r < size and alist[r] > alist[largest]):
+        largest = r
+    if (largest != index):
+        alist[largest], alist[index] = alist[index], alist[largest]
+        max_heapify(alist, largest, size)
+ 
+ 
+alist = input('Enter the list of numbers: ').split()
+alist = [int(x) for x in alist]
```
sridhar-5, line 40: There is a failing test that is pointing to this line. I think replacing the line 40 with the snippet below might solve this.

if you still have a failing test please comment.

alist = [x for x in alist]
alist = list(map(int, alist))

sridhar-5 (changes requested):
check the comments I've mentioned and do the changes accordingly. if you have trouble again. please comment I would be happy to help here.

sridhar-5:
Not really sure why test fails all the time. But personally this is the efficient way of doing this the test suggests in using `str` in place of `int`. But only `Int` has to be used. so the code looks good to me other than that. I personally feel this code is right and works.

Thanks and great job @Sranu2109

sridhar-5 (approved):
Code looks good to me. other than that failing test. if it's not a problem you can take it @ming-tsai

ming-tsai (changes requested):
Hi @Sranu2109,
Please add your code on the readme of Python

sridhar-5:
@ming-tsai ,I think the read me is now updated by @Sranu2109. can you check ?

ming-tsai (approved):

ming-tsai:
> @ming-tsai ,I think the read me is now updated by @Sranu2109. can you check ? 

Thanks

ming-tsai (changes requested):
Hi @Sranu2109,
Could you check why Python GitHub action is failed?

Sranu2109:
> Hi @Sranu2109,
> Could you check why Python GitHub action is failed?

You could use Pip-Installer from Github Actions to run package installation scripts, refer to this article [here](https://github.com/marketplace/actions/pip-installer).

It seems that the `mypy` package is missing from the build.

And/or you could add a `requirements.txt` top your project root and modify your `actions.yaml` to this.
```
steps:
- uses: actions/checkout@v2
- name: Set up Python
  uses: actions/setup-python@v2
  with:
    python-version: '3.x'
- name: Install dependencies
  run: |
    python -m pip install --upgrade pip
    pip install -r requirements.txt 
```

sridhar-5:
can you add this import statement and update here if it works?.Not sure though just give it a try.

`from typing import Callable`

algorithms/Python/sorting/heap-sort.py
```diff
@@ -0,0 +1,56 @@
...
+    if (r < size and alist[r] > alist[largest]):
+        largest = r
+    if (largest != index):
+        alist[largest], alist[index] = alist[index], alist[largest]
+        max_heapify(alist, largest, size)
+
+
+alist = input('Enter the list of numbers: ').split()
+alist = [x for x in alist] 
+alist = list(map(int, alist))
```
sridhar-5, line 41: And also try typecasting alist to int.

sridhar-5 (changes requested):

sridhar-5:
If that didn't work. update here I'll find other ways.

Arsenic-ATG (commented):
the python check was failing due to the bug in mypy where it fails to identify callable types. 

it is also reported as a bug on their repo, you can check that [here](https://github.com/python/mypy/issues/8117)

sridhar-5 (approved):
Nice Work @Sranu2109  This can be Merged @ming-tsai.

Thanks.

Arsenic-ATG:
the last 2 commits I did were minute changes which required to pass test cases, and the changes include :- 

1. Adding extra newline between import and rest of the code ( to pass isort check )

2. Adding automated test cases instead of using input() and thus removing the interactive element from the program making unit testing possible. ( to pass pytest check )

ming-tsai (approved):

sridhar-5:
Great @Arsenic-ATG.

ming-tsai:
Nice team work

PR #314, opened 2021-05-18
enh(CPlusPlus): add new functions to binary-search-tree

Ashborn-SM:

- [ ] Bugfix
- [ ] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [x] Other, please describe: Add new functions to existing algorithm

Add functions to check if a binary search tree is full and perfect.

**Other information:**


algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -47,6 +48,118 @@ void Insert(TreeNode*& root, const int& data){
...
+        }
+}
+
+int count_nodes(TreeNode* cur){
+        /**
+         * Count the number of node's in the tree
+         * @params: `cur`  node of the tree
+         */
+
+        if(cur != nullptr){ return 1 + count_nodes(cur->left) + count_nodes(cur->right); }
```
Arsenic-ATG, line 119: What if `cur` is nullptr ? 

you are lucky that gcc handles it for you by sending a zero in this case but every compiler is not gcc ( or similar ).

it's recommended to be prepare for worse and avoid letting your compiler take decision for you, and put a `return 0` here in any other case. 
```suggestion
        if(cur != nullptr){ return 1 + count_nodes(cur->left) + count_nodes(cur->right); }
        return 0;
```

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -47,6 +48,118 @@ void Insert(TreeNode*& root, const int& data){
...
+
+
+int height(TreeNode* cur){
+        /**
+         * Find the height of the tree
+         *
+         * @params: `cur`  node of the tree
+         */
+
+        if(cur != nullptr){ return 1 + std::max(height(cur->left),  height(cur->right)); }
```
Arsenic-ATG, line 130:
```suggestion
        if(cur != nullptr){ return 1 + std::max(height(cur->left),  height(cur->right)); }
        return 0;
```

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -47,6 +48,118 @@ void Insert(TreeNode*& root, const int& data){
...
+         *
+         * @return: true if the binary tree is perfect else false
+         */
+
+        if(cur == nullptr)                                { return true; }
+        if(cur->left == nullptr && cur->right == nullptr) { return depth == level; }
+        if(cur->left == nullptr || cur->right == nullptr) { return false; }
+        if(cur->left != nullptr && cur->right != nullptr) {
+                return perfect_recursive(cur->left, depth, level+1) && perfect_recursive(cur->right, depth, level+1);
+        }
```
Arsenic-ATG, line 110:
```suggestion
        if(cur == nullptr)                                { return true; }
        if(cur->left == nullptr && cur->right == nullptr) { return depth == level;
        if(cur->left != nullptr && cur->right != nullptr) {
                return perfect_recursive(cur->left, depth, level+1) && perfect_recursive(cur->right, depth, level+1);
        }
        
        return false;
```

Arsenic-ATG (changes requested):

PR #320, opened 2021-05-22
chore(CPlusPlus): add in-order predecessor and successor for bst

Ashborn-SM:
Delete in-order-Predecessor-and-successor.cpp

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Find the in-order predecessor and successor for binary search tree

**Other information:**


algorithms/CPlusPlus/Trees/in-order-predecessor-and-successor.cpp
```diff
@@ -0,0 +1,187 @@
...
+         * @params: `node` root/parent node of the tree
+	 * @params: `data` data to compare with
+	 * @params: `value` holds the previous smaller value to data
+	 *
+	 * @return: previous smallest value to data
+	 */
+	
+	if(node->data < data){ return predecessor(node->right, data, node->data); }	
+	else if(node->data > data){ return predecessor(node->left, data, value); }
+	return value;
```
Ashborn-SM, line 75: Here it will return `0` if there is no predecessor.
So if it's `0` i can print there is no predecessor but what if the predecessor is actually zero what should i do then?

algorithms/CPlusPlus/Trees/in-order-predecessor-and-successor.cpp
```diff
@@ -0,0 +1,187 @@
...
+         * @params: `root` root/parent node of the tree
+	 * @params: `data` data to compare with
+	 * @params: `value` holds the next biggest value to data
+	 *
+	 * @return: previous smaller value to data
+	 */
+
+	if(node->data < data){ return successor(node->right, data, value); }	
+	else if(node->data > data){ return successor(node->left, data, node->data); }
+	return value;
```
Ashborn-SM, line 119: the same goes for here as well

algorithms/CPlusPlus/Trees/in-order-predecessor-and-successor.cpp
Arsenic-ATG: instead of returning the value in the node, return a `TreeNode` type ( or a pointer to it ) which is holding/pointing to the required node in the tree and return NULL if there is no  predecessor/successor.

Arsenic-ATG (approved):

PR #324, opened 2021-05-25
chore(CPlusPlus): created merge-sort

NegiAkash890:

- [x] New algorithm

Added merge sort algorithm to C++ sorting folder

### DIFF
--- a/algorithms/CPlusPlus/Sorting/merge-sort.cpp
+++ b/algorithms/CPlusPlus/Sorting/merge-sort.cpp
@@ -1,5 +1,6 @@
 //Description : Merge Sort Algorithm
 #include<bits/stdc++.h>
+using namespace std;
 
 void merging(int arr[], int l, int h, int m) {
     int N1 = m - l + 1;                 //Left Array Size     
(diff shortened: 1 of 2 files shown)


algorithms/CPlusPlus/Sorting/merge-sort.cpp
```diff
@@ -0,0 +1,59 @@
+//Description : Merge Sort Algorithm
+#include<bits/stdc++.h>
+
```
aayushjain7, line 3: `using namespace std;` is missing. It will return an error.

algorithms/CPlusPlus/Sorting/merge-sort.cpp
```diff
@@ -0,0 +1,59 @@
...
+    merge_sort(arr, 0, 6);
+    cout << "Array After Sorting" << endl;
+    for (int i = 0; i < 7; i++) {
+        cout << arr[i] << " ";
+    }
+    return 0;
+   //Time Complexity :O(nlogn)
+   //Auxiliary Space :O(n)
+   //where n is the array size
+}
```
aayushjain7, line 59: Provide 1 or 2 test cases also.

aayushjain7 (changes requested):

PR #332, opened 2021-06-01
chore(CPlusPlus): add fractional-knapsack 

avanisingh0105:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Add fractional knapsack to /arrays and readme

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
+++ b/algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
@@ -1,0 +1,65 @@
+// Description : Given weights and values of n items, put these items in a knapsack of capacity W to get the maximum total value in the knapsack.
+// We can break items for maximizing the total value of knapsack/
+
+// Algorithm Type: Divide & Conquer
+// Time Complexity: O(n*log(n))
+#include <bits/stdc++.h>
+                                                                                                  using namespace std;
+class Item
+{
+public:
+    int value, weight;
+
+    // Constructor
+    Item(int value, int weight)
+    {
+        this->value = value;
+        this->weight = weight;
+    }
+};
+bool comp(Item a, Item b)
+{
+    double r1 = (double)a.value / (double)a.weight;
+    double r2 = (double)b.value / (double)b.weight;
+    return r1 > r2;
+}
+
+//Function to get the maximum total value in the knapsack.
+double fractionalKnapsack(int W, Item arr[], int n)
+{
+
+    sort(arr, arr + n, comp);
+
+    int curWeight = 0;
+    double finalvalue = 0.0;
+
+    for (int i = 0; i < n; i++)
+    {
+
+        if (curWeight + arr[i].weight <= W)
+        {
+            curWeight += arr[i].weight;
+            finalvalue += arr[i].value;
+        }
+
+        else
+        {
+            int remain = W - curWeight;
+            finalvalue += (arr[i].value / (double)arr[i].weight) * (double)remain;
+            break;
+        }
+    }
+
+    return finalvalue;
+}
+int main()
+{
+    int W = 50;                                    //    Weight of knapsack
+    Item arr[] = {{60, 10}, {100, 20}, {120, 30}}; //    {Value-Weight}pairs
+
+    int n = sizeof(arr) / sizeof(arr[0]); // size of item array
+
+    // Function call
+    cout << "Maximum value we can obtain = " << fractionalKnapsack(W, arr, n);
+    return 0;
+}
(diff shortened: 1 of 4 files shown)


avanisingh0105:
@Arsenic-ATG 
@UG-SEP 
@aayushjain7 
Can someone please review my PR?

algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    sort(arr, arr + n, comp);
+
+    int curWeight = 0;
+    double finalvalue = 0.0;
+
+    for (int i = 0; i < n; i++)
+    {
+
+        if (curWeight + arr[i].weight <= W)
```
aayushjain7, line 39: remove unnecessary spaces

algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    return finalvalue;
+}
+int main()
+{
+    int W = 50;                                    //    Weight of knapsack
+    Item arr[] = {{60, 10}, {100, 20}, {120, 30}}; //    {Value-Weight}pairs
+
```
aayushjain7, line 59: also, make it a user-defined inputs!

aayushjain7 (changes requested):

algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
Arsenic-ATG: Can you point out where exactly have he got additional spacing ? because the code looks fairly readable to me with appropriate indentation

aayushjain7: Line 30 32 35. Just the unnecessary spaces.

Arsenic-ATG: Don't get me wrong, but addition of just one line newline character to separate variable initialisation block from rest of the function body to enhance the readability looks totally ok to me.

In fact it is a good code formatting practice as it help visually separate out logical blocks of the code and reduce headaches while debugging the program later down the line. You can try using formatting tools like [clang-format](https://clang.llvm.org/docs/ClangFormat.html) or some online code formatters like [cpp formatter](https://codebeautify.org/cpp-formatter-beautifier) and see that even they approve of this.

aayushjain7: Okay, no issues! Will keep that in mind!

aayushjain7:
@avanisingh0105 have you made the changes?

avanisingh0105 (commented):
I have made the changes

aayushjain7 (dismissed):
Good work

algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+int main()
+{
+    int W; //    Weight of knapsack
+    cout << "Enter the weight(W) of the knapsack" << endl;
+    cin >> W;
+    int n;
+    cout << "Enter the total number(n) of {Value-Weight}pairs" << endl;
+    cin >> n;
+    Item arr[n];
```
Arsenic-ATG, line 63: VLA ( Variable Length Arrays) are not supported by C++ standards, you are lucky that your compiler(most probably gcc) is supporting it as an extension.

Arsenic-ATG: this not only makes your code non portable, but also make is dangerous as it opens a potential risk of having a huge array on your memory stack, making your code more venerable to stack overflow. 

It's much better to either use stl alternative ( like std::vector ) or if you are feeling brave then manually allocate ( and manage ) dynamically allocated array.

algorithms/CPlusPlus/Arrays/fractional-knapsack.cpp
```diff
@@ -0,0 +1,77 @@
+// Description : Given weights and values of n items, put these items in a knapsack of capacity W to get the maximum total value in the knapsack.
+// We can break items for maximizing the total value of knapsack/
+
+// Algorithm Type: Divide & Conquer
+// Time Complexity: O(n*log(n))
+#include <bits/stdc++.h>
+using namespace std;
+class Item
+{
+public:
```
Arsenic-ATG, line 10: If there was no private member in the class then why not go for struct at the first place ?

Arsenic-ATG (changes requested):

avanisingh0105:
Made the changes, please review.

Arsenic-ATG (approved):

aayushjain7 (approved):

PR #334, opened 2021-06-02
chore(Python): added interpolation search file

RobHam99:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Added interpolation search, in Python

**Other information:**

### DIFF
--- a/algorithms/Python/searching/interpolation_search.py
+++ b/algorithms/Python/searching/interpolation_search.py
@@ -11,18 +11,22 @@ def interpolation_search(array, x):
             min = 0
             max = len(array) - 1
             while min <= max:
-                p = int(min + (x-array[min])*(max-min) / (array[max]-array[min]))
-                if array[p] == x:
-                    return p
-                elif x < array[p]:
-                    max = p - 1
-                    continue
-                elif x > array[p]:
-                    min = p + 1
-                    continue
+                if array[min] == array[max] == x:
+                    return min, max
+                elif array[min] == array[max] != x:
+                    return False
+                else:
+                    p = int(min + (x-array[min])*(max-min) / (array[max]-array[min]))
+                    if array[p] == x:
+                        return p
+                    elif x < array[p]:
+                        max = p - 1
+                        continue
+                    elif x > array[p]:
+                        min = p + 1
+                        continue
         else:
             return False # if x is not in array return false
-
     else:
         raise Exception('Please make sure x is an integer and array is a list.')
 
--- a/algorithms/Python/searching/interpolation_search.py
+++ b/algorithms/Python/searching/interpolation_search.py
@@ -12,7 +12,7 @@ def interpolation_search(array, x):
             max = len(array) - 1
             while min <= max:
                 if array[min] == array[max] == x:
-                    return min, max
+                    return min
                 elif array[min] == array[max] != x:
                     return False
                 else:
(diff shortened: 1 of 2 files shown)


cyberwizard1001:
Looks good to me. Follows python naming conventions and algorithm works for my testcases.

cyberwizard1001:
@ming-tsai I am not a reviewer with write access.

ming-tsai:
> @ming-tsai I am not a reviewer with write access. 

That doesn't affect you review the pull request

RobHam99:
Do I need to do anything more?

cyberwizard1001 (approved):
Looks good

algorithms/Python/searching/interpolation_search.py
```diff
@@ -0,0 +1,39 @@
...
+    if isinstance(x, int) and isinstance(array, list): # check that x is integer and array is list
+        if len(array) < 1: # if array length is less than 1 print message
+            return False
+        elif len(array) == 1 and x in array: # if x is only element of array return index 0
+            return 0
+        elif len(array) > 1 and x in array: # search if array length > 1 and x is in the array
+            min = 0
+            max = len(array) - 1
+            while min <= max:
+                p = int(min + (x-array[min])*(max-min) / (array[max]-array[min]))
```
Arsenic-ATG, line 14: This formula will fail in the case when all the elements of the array are same ( i.e. `array[max]` == `array[min]` ) with  a `ZeroDivisionError`.

Arsenic-ATG (changes requested):
- [x] Handle the case where probe finding formula fails ( either by an additional conditional statement or by altering the formula itself ) 

- [x] Update the index readme file ( [DSA/algorithms/Python/README.md](https://github.com/MakeContributions/DSA/blob/main/algorithms/Python/README.md) ) of your fork with your creation.

algorithms/Python/searching/interpolation_search.py
RobHam99: I have updated this

Arsenic-ATG: Shouldn't the output be consistant ?

the output of your new special case is (`min, max`) which is inconsistent compared to any other test case where it just returns only one value at where the search value is to be found.

RobHam99: but if array[max]==array[min] then there are two elements with the target value, so i thought i should return both indices. Is it better to only return one index?

Arsenic-ATG: consider the following cases 

---
**Case 1 :** 
```
arr = [1,1,1,1,1,1]
target_value = 1
```
your function returns `(0,5)`

---
**Case 2 :** 
```
arr = [0,0,0,1,1,1,1,1,1,2,3,4,5]
target_value = 1
```
your function returns `3`

---

You can see that although the answer is correct in both the cases, it shows inconsistency in output where for one exceptional case, the function is returning more than one values and for other it isn't.

either make the function return only one value, or handle the duplicates and return all the instances of the value found in the array ( both are fairly trivial tasks )

RobHam99: Ok I have made it only return the min index, is this done now?

Arsenic-ATG (approved):

Issue #284, opened 2021-05-06
Wanted to add Interpolation Search in c


UG-SEP:
Hey @ming-tsai,
I wanted to add an interpolation search please assign me to do that.
Thanks

ming-tsai:
Hi @UG-SEP,
How is going on with this issue?

PR #337, opened 2021-06-05
chore(C): added Interpolation Search

UG-SEP:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

I have added an Interpolation search using the c language and also modified the readme.MD of C folder

**Other information:**

closes: #284

### DIFF
--- a/algorithms/C/searching/Interpolation-search.c
+++ b/algorithms/C/searching/Interpolation-search.c
@@ -8,7 +8,7 @@ Interpolation search: Used to search data in a sequence stored array
 #include <stdlib.h>
  // Function to search data in a given array
 int interpolation_search(int * arr, int low, int high, int to_search) {
-  if (low <= high && to_search >= arr[low] && to_search <= arr[high]) {
+  if (low <= high && to_search >= arr[low]) {
         if(arr[high]==arr[low]&&to_search==arr[high]&&arr[low]==to_search)
         return high;
     // formula to find index
--- a/algorithms/C/searching/Interpolation-search.c
+++ b/algorithms/C/searching/Interpolation-search.c
@@ -1,0 +1,69 @@
+/*
+Program to search data in array using interpolation search
+Interpolation search: Used to search data in a sequence stored array
+*/
+
+#include<stdio.h>
+#include<stdlib.h>
+// Function to search data
+int interpolation_search(int *arr,int low,int high,int to_search)
+{
+    if (low<=high&&to_search>=arr[low]&&to_search<= arr[high])
+    {
+    // formula to find index
+int pos=low+(to_search-arr[low])*(high-low)/(arr[high]-arr[low]);
+
+    // if data found
+    if(arr[pos]==to_search)
+        return pos;
+    // if data is less than current position
+    else if(arr[pos]>to_search)
+        high=pos-1;
+    // if data is greater than the current position
+    else if(arr[pos]<to_search)
+        low=pos+1;
+    // recursive call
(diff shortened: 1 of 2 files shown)


algorithms/C/searching/Interpolation-search.c
```diff
@@ -0,0 +1,69 @@
...
+    int *arr,size,index,to_search,i;
+
+    printf("Enter the size: ");
+    // taking size
+    scanf("%d",&size);
+    // dynamically locating memory
+    arr=(int*)malloc(sizeof(int)*size);
+    for(i=0;i<size;i++)
+    {
+        scanf("%d",&arr[i]);
```
Arsenic-ATG, line 43: **Possible null dereference [CWE-690]:** 

`arr` will point to `NULL` in case malloc() fails to allocate the memory ( maybe due to insufficient memory or memory corruption ) 

**Fix:**
perform a check before entering the loop for allocation failure.

algorithms/C/searching/Interpolation-search.c
```diff
@@ -0,0 +1,69 @@
...
+
+#include<stdio.h>
+#include<stdlib.h>
+// Function to search data
+int interpolation_search(int *arr,int low,int high,int to_search)
+{
+    if (low<=high&&to_search>=arr[low]&&to_search<= arr[high])
+    {
+    // formula to find index
+int pos=low+(to_search-arr[low])*(high-low)/(arr[high]-arr[low]);
```
Arsenic-ATG, line 14: **Possible Division by zero [CWE-369] :** 

In case the array is filled with same values as that of target ( `arr[high]` == `arr[low]` ) the denominator of the probe calculation formula will become zero.

**Fix:**
either reconstruct the probe calculating formula or handle this edge case explicitly.

Arsenic-ATG (changes requested):
- [ ] Prefer fixing spacing and indents in your program to make it more pleasing to eyes ( if you are feeling lazy then you can also do this with the help of some automation tools like [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) or online one like [code beautify](https://codebeautify.org/cpp-formatter-beautifier)

- [ ] fix the possible runtime errors pointed out in comments 

- [ ] If you are creating an UI and making it possible for user to input stuff then prefer separating different output statements with a newline character ( again to make is easier on eyes even if the user is using non interactive shells or automated testing environment )

UG-SEP:
@Arsenic-ATG hope this solves the problem
Guide me if I am wrong

algorithms/C/searching/Interpolation-search.c
```diff
@@ -0,0 +1,72 @@
...
+
+#include <stdio.h>
+
+#include <stdlib.h>
+ // Function to search data in a given array
+int interpolation_search(int * arr, int low, int high, int to_search) {
+  if (low <= high && to_search >= arr[low] && to_search <= arr[high]) {
+        if(arr[high]==arr[low]&&to_search==arr[high]&&arr[low]==to_search)
+        return high;
+    // formula to find index
```
Arsenic-ATG, line 14: No need to perform extra comparison as the third condition will be a tautology if the first 2 are true.

```suggestion
  if (low <= high && to_search >= arr[low] && to_search <= arr[high]) {
      if( arr[high] == arr[low] && to_search==arr[high] )
        return high;
    // formula to find index
```

Arsenic-ATG (approved):

PR #343, opened 2021-06-07
chore(CPlusPlus): add min-heap

Ashborn-SM:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Add min heap structure

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/README.md
+++ b/algorithms/CPlusPlus/README.md
@@ -67,6 +67,7 @@
 6. [In order morris traversal](Trees/in-order-morris-traversal.cpp)
 7. [In order Predecessor and Successor](Trees/in-order-predecessor-and-successor.cpp)
 8. [Avl Tree](Trees/avl.cpp)
+9. [Min Heap](Trees/min-heap.cpp)
 
 # Maths
 1. [Kaprekar Number](Maths/Kaprekar-number.cpp)
--- a/algorithms/CPlusPlus/Trees/min-heap.cpp
+++ b/algorithms/CPlusPlus/Trees/min-heap.cpp
@@ -7,7 +7,6 @@
 */
 
 #include <iostream>
-#include <stdexcept>
 
 class Heap{
 	public:
@@ -58,7 +57,7 @@ void Heap::growArray(){
 	// Doubling the capacity
 	capacity *= 2;
 	// delete the items inorder to avoid any memory leak
-	delete items;
+	delete[] items;
 	// set items to point to new_Array
 	items = new_Array;
 }
--- a/algorithms/CPlusPlus/Trees/min-heap.cpp
+++ b/algorithms/CPlusPlus/Trees/min-heap.cpp
@@ -1,0 +1,139 @@
+/*
+ Implementation of the min-heap using an array.
+
+ Min-Heap is a Binary tree where each child nodes value is greater 
+ than that of their parent. It is used when you need quick access to 
+ the smallest number in the array.
+*/
+
+#include <iostream>
+#include <stdexcept>
+
+class Heap{
+	public:
+		void insert(const int& data);
+		void removemin();
+		void print();
+		Heap();
+		~Heap();
+
+	private:
+		// size - number of elements in the array
+		// capacity - number of elements the array can store
+		int size, capacity;
+		int *items;
+
+		int parent(unsigned index) const;
+		int child(unsigned index) const;
+
+		void heapifyUP(unsigned index);
+		void heapifyDOWN(unsigned index);
+		void growArray();
+
+};
+
+Heap::~Heap(){
+	delete[] items;
+	items = nullptr;
+}
+
+Heap::Heap(): size(0), capacity(2){
+	items = new int[capacity + 1];
+}
+
+int Heap::parent(unsigned index) const{ return index/2; }
+
+void Heap::growArray(){
+	/**
+	  * Inorder to accommodate more numbers in the array
+	  * Allocate memory in the heap 
+	  * Double the size of the old array up to (capacity*2) + 1
+	  * Copy the elements of the previous array into the new array
+	  */
+
+	// Doubling the size of the array
+	int* new_Array = new int[(capacity * 2) + 1];
+	// Copying the elements of the old array to the new array
+	for(int i=1; i<=size; i++){ new_Array[i] = items[i]; }
+	// Doubling the capacity
+	capacity *= 2;
+	// delete the items inorder to avoid any memory leak
+	delete items;
+	// set items to point to new_Array
+	items = new_Array;
+}
+
+void Heap::insert(const int& data){
+	// if size == capacity it means the array is full and need to grow
+	if(size == capacity){ growArray(); }
+	items[++size] = data;
+	heapifyUP(size);
+}
+
+void Heap::heapifyUP(unsigned index){
+	if(index > 1){
+		if(items[index] < items[parent(index)]){
+			std::swap(items[index], items[parent(index)]);
+			heapifyUP(parent(index));
+		}
+	}
+}
+
+void Heap::removemin(){
+	std::swap(items[1], items[size--]);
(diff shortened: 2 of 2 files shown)


algorithms/CPlusPlus/README.md
```diff
@@ -67,6 +67,7 @@
 6. [In order morris traversal](Trees/in-order-morris-traversal.cpp)
 7. [In order Predecessor and Successor](Trees/in-order-predecessor-and-successor.cpp)
 8. [Avl Tree](Trees/avl.cpp)
+9. [Min-Heap](Trees/min_heap.cpp)
```
UG-SEP, line 70:
```suggestion
9. [Min-Heap](Trees/min-heap.cpp)
```
Please read the [readme.md](https://github.com/MakeContributions/DSA/blob/main/README.md) Contributing Specifications that you have to separate word with `-` not `_`

algorithms/CPlusPlus/Trees/min_heap.cpp
```diff
@@ -0,0 +1,109 @@
+#include <iostream>
```
UG-SEP, line 1: add a user-friendly description about the program
`hints: Implementation of the min-heap using an array.`
`Min-Heap: Minimum Heap is a method of arranging elements in a binary search tree where the value of the parent node is lesser than that of its child nodes.`

algorithms/CPlusPlus/Trees/min_heap.cpp
```diff
@@ -0,0 +1,109 @@
...
+	private:
+		int size_, capacity_;
+		int *items_;
+
+		int _parent(unsigned index) const;
+		int _child(unsigned index) const;
+
+		void _heapifyUP(unsigned index);
+		void _heapifyDOWN(unsigned index);
+		void _growArray();
```
UG-SEP, line 20:
```suggestion
		int size, capacity;
		int *items;

		int parent(unsigned index) const;
		int child(unsigned index) const;

		void heapifyUP(unsigned index);
		void heapifyDOWN(unsigned index);
		void growArray();
```
using `_` as first char in variable name looks ugly

algorithms/CPlusPlus/Trees/min_heap.cpp
```diff
@@ -0,0 +1,109 @@
...
+	for(int i=1; i<=size_; i++){ new_Array[i] = items_[i]; }
+	capacity_ *= 2;
+	items_ = new_Array;
+}
+
+void Heap::insert(const int& data){
+	if(size_ == capacity_){ _growArray(); }
+	items_[++size_] = data;
+	_heapifyUP(size_);
+}
```
UG-SEP, line 41: add comments where hard to understand

UG-SEP (changes requested):
Please do the below-given changes in your code

algorithms/CPlusPlus/README.md
Ashborn-SM: My bad

algorithms/CPlusPlus/Trees/min_heap.cpp
Ashborn-SM: isn't `_` the naming convention for private variable in c++?

Arsenic-ATG: @Ashborn-SM looks like you are from C# background where people generally prefer using underscore as a prefix ( or suffix ) for private members of the the class to easily spot the scope of the variable when used and somewhat enhance the readability of the program.

But for C++, the case is different and can be quite disastrous especially when you use underscore as prefix while naming your identifier. It's because for C/C++, such names are reserved for language implementation to use and using it inside your code can cause conflicts with implementation variables leading to unwanted bugs which can be very hard to debug.

Ashborn-SM: thanks @Arsenic-ATG,
i will do the necessary changes

algorithms/CPlusPlus/README.md
```diff
@@ -67,6 +67,7 @@
 6. [In order morris traversal](Trees/in-order-morris-traversal.cpp)
 7. [In order Predecessor and Successor](Trees/in-order-predecessor-and-successor.cpp)
 8. [Avl Tree](Trees/avl.cpp)
+9. [Min-Heap](Trees/min-heap.cpp)
```
UG-SEP, line 70:
```suggestion
9. [Min Heap](Trees/min-heap.cpp)
```

algorithms/CPlusPlus/Trees/min-heap.cpp
```diff
@@ -0,0 +1,133 @@
...
+		void insert(const int& data);
+		void removemin();
+		void print();
+		Heap();
+
+	private:
+		// size - number of elements in the array
+		// capacity - number of elements the array can store
+		int size, capacity;
+		int *items_;
```
UG-SEP, line 23:
```suggestion
              int *items;
```

UG-SEP (changes requested):
Please do the changes

UG-SEP (dismissed):
Thanks a lot, @Ashborn-SM for giving your precious time to contribute to the DSA repo will we wait for your next PR😊😊

algorithms/CPlusPlus/Trees/min-heap.cpp
```diff
@@ -0,0 +1,133 @@
...
+	/*
+	Output: 4 10 18 13 42 22 45 95
+	
+	Time complexity to build the heap: O(n)
+	Time complexity to remove min: O(log(n))
+	Time complexity to remove all elements: O(n*log(n))
+	*/
+	return 0;
+
+}
```
Arsenic-ATG, line 133: **`heap.Heap::items` will leak here [[CWE-401]](https://cwe.mitre.org/data/definitions/401.html) :** 

I can't see you deallocating the the memory you allocated dynamically ( in constructor of object `heap` ), and as soon as the object goes out of scope, it will be destroyed, causing a memory leak.

Fix: 
delete the memory allocated to the array of the object as soon as it goes out of scope. Best place to do this would be in the class destructor of your data structure which is meant to perform these kind of cleanup jobs for the structure.

Arsenic-ATG (changes requested):

algorithms/CPlusPlus/Trees/min-heap.cpp
```diff
@@ -0,0 +1,139 @@
...
+		int child(unsigned index) const;
+
+		void heapifyUP(unsigned index);
+		void heapifyDOWN(unsigned index);
+		void growArray();
+
+};
+
+Heap::~Heap(){
+	delete items;
```
Arsenic-ATG, line 36:
```suggestion
	delete[] items;
```

algorithms/CPlusPlus/Trees/min-heap.cpp
```diff
@@ -0,0 +1,139 @@
...
+	  */
+
+	// Doubling the size of the array
+	int* new_Array = new int[(capacity * 2) + 1];
+	// Copying the elements of the old array to the new array
+	for(int i=1; i<=size; i++){ new_Array[i] = items[i]; }
+	// Doubling the capacity
+	capacity *= 2;
+	// delete the items inorder to avoid any memory leak
+	delete items;
```
Arsenic-ATG, line 61:
```suggestion
	delete[] items;
```

Arsenic-ATG (changes requested):

algorithms/CPlusPlus/Trees/min-heap.cpp
Arsenic-ATG: looks like you missed the deallocation in `void Heap::growArray()` function

Ashborn-SM: All done!!

Arsenic-ATG (approved):

PR #353, opened 2021-06-12
enh(CPlusPlus): circular linked lists

NegiAkash890:
This refactors the previous code to make it more readable and user friendly

- [ ] Bugfix
- [ ] New algorithm
- [x] Code style update
- [x] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

Changes are still not final, I will make the changes soon

### DIFF
--- a/algorithms/CPlusPlus/Linked-Lists/circular.cpp
+++ b/algorithms/CPlusPlus/Linked-Lists/circular.cpp
@@ -1,39 +1,56 @@
 #include <iostream>
 using namespace std;
-struct Node {
-   int data;
-   struct Node *next;
+struct Node
+{
+  int data;
+  Node *next;
+  Node (int x)
+  {
+    data = x;
+    next = NULL;
+  }
 };
-struct Node* head = NULL;
-void insert(int newdata) {
-   struct Node *newnode = (struct Node *)malloc(sizeof(struct Node));
-   struct Node *ptr = head;
-   newnode->data = newdata;
-   newnode->next = head;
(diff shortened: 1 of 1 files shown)


algorithms/CPlusPlus/Linked-Lists/circular.cpp
```diff
@@ -1,39 +1,56 @@
...
-� � � ptr = ptr->next;
-� � � ptr->next = newnode;
-� �} else
-� �newnode->next = newnode;
-� �head = newnode;
+Node *insertion (int newdata,Node *head)
+{
+  Node *newnode = new Node(newdata);
+  Node *ptr = head;
+  if(head == NULL){
```
Arsenic-ATG, line 17: Q what will `head->next` point to in this case ?

UG-SEP:
## Warning❗❗
@anirudh09041  if you do not respond in 2 weeks then we will close the PR

Arsenic-ATG (approved):
functionally works

but could have been better, as in current state it is only suitable for one small list with very limited use.

some of the upgrades include :- 

1. encapsulating the entire data structure under one class so that can be used more elegantly and user don't have to do something like `head = insertion(4, head);`

2. fixing the memory leak that the program generates by not deallocating the all the memory allocated to the list which.

3. adding more functionality to the structure itself as currently it is just a showcase in which you can place your data for the sake of seeing it.

but these are the things which can be improved in any future update by same of other devs/ contributors.

PR #359, opened 2021-06-16
chore(Python): add knapsack problem

codelight2711:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/Python/dynamic_programming/0 1 Knapsack .py
+++ b/algorithms/Python/dynamic_programming/0 1 Knapsack .py
@@ -1,0 +1,43 @@
+'''
+0 1 Knapsack - Problem
+Send Feedback A thief robbing a store can carry a maximal weight of W into his knapsack. There are N items, and i-th item weigh 'Wi' and the value being 'Vi.' What would be the maximum value V, that the thief can steal?
+'''
+
+
+from sys import stdin
+
+def knapsack(weights, values, n, maxWeight) :
+    K=[[0 for x in range(maxWeight+1)] for x in range(n+1)]
+    for i in range(n+1):
+        for w in range(maxWeight+1):
+            if i==0 or w==0:
+                K[i][w]=0
+            elif weights[i-1]<=w:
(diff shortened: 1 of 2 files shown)


codelight2711:
what is the issue, why is this failing ?

ming-tsai:
Hi @codelight2711,
The Python check is failing, see the detail
image
image

algorithms/Python/dynamic_programming/0 1 Knapsack .py
```diff
@@ -0,0 +1,34 @@
...
+    for i in range(n+1):
+        for w in range(maxWeight+1):
+            if i==0 or w==0:
+                K[i][w]=0
+            elif weights[i-1]<=w:
+                K[i][w]=max(values[i-1]+K[i-1][w-weights[i-1]],K[i-1][w])
+            else:
+                K[i][w]=K[i-1][w]
+    return K[n][w]
+def takeInput() :
```
Arsenic-ATG, line 18: Instead of taking input, it's preferred to either include a doctest or run the program on a predefined custom test-case/cases. It will make the automated unit testing process a bit faster and easier.

Arsenic-ATG (changes requested):
sorry for late response, I was quite busy this week. 

1 ) **Regarding the failing check :** It is because of `isort` which checks the formatting of the program, and in this case, it is raising error because according to PEP8, you have to surround top level functions with 2 blank lines and it can't see that in your programs. 

2 ) apart from that these are the things that needs fixing :- 
- [ ] update [Readme.md](https://github.com/MakeContributions/DSA/blob/main/algorithms/Python/README.md) of python repo with your creations under appropriate topics
- [ ] avoid using whitespaces in the names of the file ( prefer using a udnerscore "_" instead)
- [ ] instead of taking input from the user, prefer automated test ( via predefined test cases inside the program )

hope this clears all the queries, but if still you have any query then feel free to ask here.

UG-SEP:
## Warning❗❗
@codelight2711 if you do not respond in 2 weeks then we will close the PR

PR #376, opened 2021-07-01
chore(C): add binary search tree

Ashborn-SM:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Add binary search tree using c language

**Other information:**

### DIFF
--- a/algorithms/C/tree/binary-search-tree.c
+++ b/algorithms/C/tree/binary-search-tree.c
@@ -72,6 +72,14 @@ void print(struct node* root){
 	}
 }
 
+void deallocate(struct node** root){
+	if(*root != NULL){
+		deallocate(&(*root)->left);
+		deallocate(&(*root)->right);
+		free(*root);
+		*root = NULL;
+	}
+}
 
 int main(){
 	struct node* root = NULL;
@@ -110,5 +118,6 @@ int main(){
 			BST is Perfect
 			4 19 22 37 42 51 55
 	*/
+	deallocate(&root);
 	return 0;
 }
--- a/algorithms/C/tree/binary-search-tree.c
+++ b/algorithms/C/tree/binary-search-tree.c
@@ -1,0 +1,114 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+
+struct node{
+	struct node* left;
+	struct node* right;
+	int val;
+};
+
+struct node* createNode(int val){
+	struct node* n = (struct node*)malloc(sizeof(struct node));
+	n->val = val;
+	n->left = NULL;
+	n->right = NULL;
+	return n;
+}
+
+struct node* insert(struct node* root, int val){
+	if(root == NULL){
+		root = createNode(val);
+	}
+	else if(root->val < val){
+		root->right = insert(root->right, val);
+	}
+	else{
+		root->left = insert(root->left, val);
+	}
+
+	return root;
+}
+
+bool is_full(struct node* root){
+	if(root == NULL || (root->left == NULL && root->right == NULL)){
+		return true;
+	}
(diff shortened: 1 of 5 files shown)


algorithms/C/tree/binary-search-tree.c
```diff
@@ -0,0 +1,123 @@
...
+	    / \  / \
+	   4  22 42 55
+
+		Output:
+			BST is full
+			BST is Perfect
+			4 19 22 37 42 51 55
+	*/
+	deallocate(root);
+	printf("\n%i", root->val);
```
Ashborn-SM, line 121: Even after i de-allocate the memory, i can access the value, why is tht?

UG-SEP: Read this [Accessing data after memory has been released](https://stackoverflow.com/questions/42588482/c-accessing-data-after-memory-has-been-freeed) this may help you...

Arsenic-ATG: @Ashborn-SM that's because, deallocating memory just makes the memory block ( that your pointer is pointing to  ) available for program or OS to allocate it to someplace else, you pointer may still point to same memory location that it was pointing before, but accessing such memory location leads to [Undefined Behaviour](https://en.wikipedia.org/wiki/Undefined_behavior). Such pointers are famously known as [dangling pointers](https://en.wikipedia.org/wiki/Dangling_pointer)

Ashborn-SM: @Arsenic-ATG @UG-SEP thanks for the help, i understand how the memory is being re-used for some other purpose by the OS but when i try to set each tree node to NULL it is not giving me a segfault, why is that? shouldn't that be the case?

UG-SEP: this is a segment fault statement you may access the root because you are changing the root from deallocating function and in that function, you pass the value of root, not the address so the change you do in root there will not change it from the main function
Hope you understand

Arsenic-ATG: > @Arsenic-ATG @UG-SEP thanks for the help, i understand how the memory is being re-used for some other purpose by the OS but when i try to set each tree node to NULL and try to access that it is not giving me a segfault, why is that? i mean shouldn't that be the case a segfault ?

not always 
A lot of things happening here are not defined by language standards, so it's either your OS or your language implementation which is deciding what should happen in such cases, in case of a segfault, it is mostly generated by your OS memory manager when a process tries to access an invalid or illegal memory address, and accessing a pointer after deallocating the memory it pointes to most likely will not generate such error since the pointer is still pointing to some real memory ( again it's not guaranteed and largely depends on how your implementation and OS manages memory ).

If you would have set the `root` node to `NULL` after deallocating it, then it would have most likely caused a segfault as generally NULL doesn't map to a valid memory location ( again exceptions do exist where it is a valid memory address like in [x86-real mode](https://wiki.osdev.org/Real_Mode) ).

Ashborn-SM: Thanks alot @Arsenic-ATG @UG-SEP. 

I am all done here, is there any changes i should i do?

Ashborn-SM: Thanks alot @Arsenic-ATG @UG-SEP.
I was passing the `root` by value ig thts why i couldn't set it to NULL.

I am all done here.

Arsenic-ATG (approved):

algorithms/C/tree/binary-search-tree.c
UG-SEP: @Ashborn-SM that great
Happy to help you 😊😊😊

PR #382, opened 2021-07-06
chore(CPlusPlus): added wave-sort

Janhavi-2001:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**


wave-sort.cpp
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+}
+
+int main()
+{
+    cout << "Enter the number of elements: ";
+    int n;
+    cin >> n;
+    int arr[n];
```
Arsenic-ATG, line 29: No VLAs[variable size arrays] please. 
these things are prohibited in C++ and using this here makes your program dependent on language implementation to have a non-standard extension to support this.

it's much better to use some `stl` containers ( like `std::vector`) for such job, or if you are feeling brave then dynamically allocate the memory to the array here ( don't forget to manage it's ownerships and deallocate it if you choose this way )

wave-sort.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+
+void swap(int arr[], int i,int j)
+{
+    int temp=arr[i];
+    arr[i]=arr[j];
+    arr[j]=temp;
+}
+
+void wavesort(int arr[], int n)
```
Arsenic-ATG, line 58: It's nice to have multiple implantation of same algorithm, but instead of commenting entire program, I think it's better to just have a function with other name ( something like `WaveSort2()` following your convention ) and call both on 2 different arrays in `main()`.

Arsenic-ATG (changes requested):
The program right now is in the root, and should be moved to corresponding correct place instead. 
I think correct place for this should be `DSA/algorithms/CPlusPlus/Sorting/`, and also don't forget to update the `readme.md` file of that dir with your creation in the list. ( details can be found on [contribution guidelines](https://github.com/MakeContributions/DSA#contribution-guidelines) of the repo 
)

Arsenic-ATG (approved):

PR #387, opened 2021-07-13
chore(CPlusPlus): add pascal triangle

Janhavi-2001:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

This algorithm is known as the Pascal's triangle algorithm, which is theoretically, a triangular array of binomial coefficients. It can also be said to be the powers of a the number 11 arranged in a triangular format.

**Other information:**
This program is iterative and user-driven i.e. the user can decide the number of coefficients to be displayed on the console as output.


algorithms/CPlusPlus/Sorting/wave-sort.cpp
```diff
@@ -0,0 +1,44 @@
+#include <iostream>
```
Arsenic-ATG, line 1: These files should not exist, looks like your fork isn't updated. ( you may check [these docs](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) if you don't know how to do it )

algorithms/CPlusPlus/Maths/pascal-triangle.cpp
```diff
@@ -0,0 +1,39 @@
...
+        if(choice == 'Y' || choice == 'y')
+        {
+            cout << "Enter the desired number till which you want to continue: ";
+            int n;
+            cin >> n;
+            for(int i=0;i<n;i++)
+            {
+                for(int j=0;j<n;j++)
+                {
+                    cout << pow(11,i);
```
Arsenic-ATG, line 23: This is a nice trick to get the nth row of the triangle but will it work for any row for n > 5 ?
 
Given the fact that C++ gives limited size to it's data types and there is no primitive datatype available to hold this huge number ( I don't think even `long long` can hold any such number ), I think it would be a better to either use a larger non-primitive data type ( like arrays ) to hold such numbers in your own representation 
or skip all that hassle and print the numbers without storing them anywhere ( by breaking down the bigger number into series of smaller numbers and printing the series instead ), something like this : -

```
1,  
1, 1 
1, 2, 1 
1, 3, 3, 1 
1, 4, 6, 4, 1 
1, 5, 10, 10, 5, 1 
```

Arsenic-ATG (changes requested):
sorry for late review, I was a bit busy with other projects. 

I have commented the changes that need to be done as review comments, please feel free to ask any query. 👍

Janhavi-2001:
Sure I'll update you with the changes soon!

ming-tsai:
Hi @Janhavi-2001,
This pull request will be closed, any update please re-open it

PR #400, opened 2021-07-27
chore(CPlusPlus): add middle-node

emily876:


UG-SEP:
Sorry but I found that your code is plagiarism from the GeeksforGeeks website [link](https://www.geeksforgeeks.org/write-a-c-function-to-print-the-middle-of-the-linked-list/)
Because the way of writing code as well as variable, comments, function name, class name everything is same
So sorry for that but we don't accept these things.
Recommendation: `Please don't plagiarism code because it down your reputation`

PR #428, opened 2021-08-25
chore(Python): added is unique character algorithm

Dhee2211:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Added  algorithm to check whether the given string has unique characters or not

Algorithm is written in python language. 

**Other information:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,4 @@ Thumbs.db
 *.idea
 
 
-algorithms/Python/arrays/unique_character.py
+
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ Thumbs.db
 *.class
 *.idea
 
+
+algorithms/Python/arrays/unique_character.py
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -12,3 +12,5 @@ Thumbs.db
 *.class
 *.idea
 
+
+algorithms/Python/arrays/unique_character.py
```
Arsenic-ATG, line 16: looks like you accidentally created this project somewhere else and didn't wanted git to track that particular file. 

it's recommended for `.gitignore` to only contain files which are unnecessary and could be generated when working with the project itself (for example: compiler generated files, project files etc. ). 

If you want some files to be untracked exclusively on your local git and don't want to pollute `.gitignore` then it's recommended to update `.git/info/exclude` instead.

Arsenic-ATG (changes requested):
The program works correctly.

Just remove the unnecessary content from `.gitignore` and update the [index README.md file](https://github.com/Dhee2211/DSA/blob/dhee2211-linked-list-examples/algorithms/Python/README.md) with your creation to make this eligible for the merging.

.gitignore
Dhee2211: I will update shortly. Thank you for the review.

Dhee2211 (commented):
Updated  .gitignore and README.md

Arsenic-ATG (approved):

PR #434, opened 2021-08-29
enh(CPlusPlus): Binary Search Tree and Iterative Traversal

shinchancode:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [x] Other, please describe:

1. I have added iterative method also for traversal of tree using stak for better understanding.
2. I have added three functions in Binary search tree : finding maximum, minimum value in a BST and printing all lead nodes present in the tree.

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Trees/binary-search-tree.cpp
+++ b/algorithms/CPlusPlus/Trees/binary-search-tree.cpp
@@ -184,7 +184,7 @@ void leaf_nodes(TreeNode* root) //Print all leafnode in BST
        leaf_nodes(root->right); 
 } 
 
-void f_min(TreeNode* root) //Find Minimum element from root
+TreeNode* f_min(TreeNode* root) //Find Minimum element from root
 { 
     /**
      * Print the minimum value of the tree
@@ -195,16 +195,17 @@ void f_min(TreeNode* root) //Find Minimum element from root
     if(root==NULL)
     {
         std::cout<<"No value present in the tree"<< std::endl;
+        return NULL;
     }
-   TreeNode* p=root; 
-   while(p->left!=NULL) 
-   { 
-       p=p->left; 
-   } 
-   std::cout<<"Minimum Value Present in the tree is : "<<p->data<< std::endl;
+    TreeNode* p=root; 
+    while(p->left!=NULL) 
+    { 
+        p=p->left; 
+    } 
+    return p;
 } 
 
-void f_max(TreeNode* root) //Find Maximum element from root
+TreeNode* f_max(TreeNode* root) //Find Maximum element from root
 { 
     /**
      * Print the maximum value of the tree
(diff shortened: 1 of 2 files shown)


algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -158,83 +159,155 @@ bool isperfect(TreeNode* root){
...
+   } 
+
+   if (root->left) 
+       leaf_nodes(root->left); 
+
+   if (root->right) 
+       leaf_nodes(root->right); 
+} 
+
+void f_min(TreeNode* root) //Find Minimum element from root
```
Arsenic-ATG, line 187: **Suggestion :**
in are general used case, one would rather use this function to get the minimum value in the tree and use it somewhere else. 

So isn't it better for this function ( and `f_max ()` ) to return the minimum value to the caller and let the caller use it for whatever it want the minimum value for ?

shinchancode: Ok sir, I will work on it. I will try to add delete function. There we need minimum value.
Thank you for suggestion

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -158,83 +159,226 @@ bool isperfect(TreeNode* root){
...
+
+    if(isperfect(root)) { std::cout << "The binary tree is PERFECT" << std::endl; }
+    else           { std::cout << "The binary tree is not PERFECT" << std::endl; }
+
+    std::cout << "Leaf Nodes present in the binary tree are : ";
+    leaf_nodes(root);
+    std::cout<< std::endl;
+    
+    n = f_max(root);
+    if(f_max(root)!=NULL)
```
Arsenic-ATG, line 333: We already have `n` holding the result of `f_max(root)`, so this function call overhead can be avoided.

```suggestion
    if(n)
```

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
```diff
@@ -158,83 +159,226 @@ bool isperfect(TreeNode* root){
...
+    {
+        std::cout<<"Maximum Value Present in the tree is : "<<n->data<< std::endl;
+    }
+
+    n=f_min(root);
+    if(f_max(root)!=NULL)
```
Arsenic-ATG, line 339: Same here
```suggestion
    if(n)
```

algorithms/CPlusPlus/Trees/binary-search-tree.cpp
Arsenic-ATG: It's nice that you also added a used case in the program also.

Arsenic-ATG (approved):

PR #468, opened 2021-09-16
chore(CPlusPlus): added Huffman

KareimGazer:
Signed-off-by: Kareim Tarek AbdelAzeem <kareimgazar1990@gmail.com>

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

added Huffman Code Algorithm in C++ with example tests

**Other information:**


algorithms/CPlusPlus/Codes/huffmanCode/README.md
```diff
@@ -0,0 +1,18 @@
+# Huffman Code
+
+## Goal
+The goal is minimize the size of files using the fact that some characters are redundent and with higher frequency than others like the word the, so the goal is to use less chars to type the most common words and more chars for the least common words.
```
Arsenic-ATG, line 4: Typo ( spotted by codespell )
```suggestion
The goal is minimize the size of files using the fact that some characters are redundant and with higher frequency than others like the word the, so the goal is to use less chars to type the most common words and more chars for the least common words.
```

algorithms/CPlusPlus/Codes/huffmanCode/huffman.h
```diff
@@ -0,0 +1,33 @@
+#include <string>
```
Arsenic-ATG, line 1: It's recommend to add a [header guard](https://en.wikipedia.org/wiki/Include_guard) to your header file.

Arsenic-ATG (changes requested):
According to me, It would be better to transfer the `algorithms/CPlusPlus/codes/huffmanCode` directory to `algorithms/CPlusPlus/Tree/huffmanCode`.

algorithms/CPlusPlus/Codes/huffmanCode/huffman.h
Arsenic-ATG: this is also the reason why `cpplint` check is failing.

KareimGazer:
@ming-tsai @Arsenic-ATG 
Fixed type and added header guard
Thanks for your patience I was busy the last couple of days.

Arsenic-ATG:
so the code is failing [`cpplint`](https://github.com/cpplint/cpplint) tests and these are the "styling faults" that it points out in your program : -

1. > algorithms/CPlusPlus/Codes/huffmanCode/huffman.cpp:10:  <regex> is an unapproved C++11 header.  [build/c++11] [5]

This is because cpplint seem to prefer an external library within Google over them.
learn more about it [here](https://github.com/google/styleguide/issues/194)

2. > algorithms/CPlusPlus/Codes/huffmanCode/huffman.h:1:  #ifndef header guard has wrong style, please use: ALGORITHMS_CPLUSPLUS_CODES_HUFFMANCODE_HUFFMAN_H_  [build/header_guard] [5]


3. > algorithms/CPlusPlus/Codes/huffmanCode/huffman.h:36:  #endif line should be "#endif  // ALGORITHMS_CPLUSPLUS_CODES_HUFFMANCODE_HUFFMAN_H_"  [build/header_guard] [5]


4. > algorithms/CPlusPlus/Codes/huffmanCode/main.cpp:1:  Include the directory when naming .h files  [build/include_subdir] [4]

a quick fix for point 4 can be to update `#include "huffman.h"` to `#include "./huffman.h"

Arsenic-ATG (approved):
My commits should have fixed the cpplint erorrs. 

Although the code still needs a makefile to be built correctly, but that can be added later, and this PR should be good to go for now.

PR #486, opened 2021-09-26
added new array program

samughodake:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Arrays/segregate0and1.cpp
+++ b/algorithms/CPlusPlus/Arrays/segregate0and1.cpp
@@ -1,0 +1,48 @@
+/*
+Description: A program to segregate the array of 0s and 1s
+
+Approach: Counting the number of 0s present in the array and saving it in variable z. For the first z indexes, putting 0 and for the remaining indexes puting 1.
+
+Time complexity: O(n)
+*/
+
+#include <iostream>
+using namespace std;
+
+void segregate0and1(int arr[], int n) {
+    
+    int z=0;
+    for(int i=0;i<n;i++){
+      //counting the number of 0s and storing it the variable z
+      if(arr[i]==0){
+        z++;
+      }
+    }
+    //for z indexes, putting 0 in the array
+    //for remaining indexes putting 1 in the array
+    for(int j=0;j<n;j++){
+      if(j<z){
+               arr[j]=0;
+      }
+      else{
+               arr[j]=1;
+      }
+    }       
+}
+int main() {
+    cout << "Enter number of elements in array: \n";
+    int n;
+    cin>>n;
+
+    cout<<"Enter the array elements (only 0 and 1)\n";
+    int arr[n];
+    for(int i=0;i<n;i++){
+      cin>>arr[i];
+    }
+    //calling the function
+    segregate0and1(arr,n);
+    cout<<"After segregating, the array is: \n";
+    for(int j=0;j<n;j++){
+      cout<<arr[j]<<" ";
+    }
+}


Ashborn-SM (changes requested):
@samughodake please do the changes that are mentioned.

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
```diff
@@ -0,0 +1,48 @@
+/*
+Description: A program to segregate the array of 0s and 1s
+
+Approach: Counting the number of 0s present in the array and saving it in variable z. For the first z indexes, putting 0 and for the remaining indexes putting 1.
```
Ashborn-SM, line 4: It would be readable if you use multiple lines instead of one single line to address this comment,
so no one has to drag their mouse to the end or just move the scroll bar horizontally just to read it. 
```suggestion
Approach: Counting the number of 0s present in the array and saving 
it in variable z. For the first z indexes, putting 0 and for the remaining 
indexes putting 1.
```

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }       
+}
+int main() {
+    cout << "Enter number of elements in array: \n";
+    int n;
+    cin>>n;
+
+    cout<<"Enter the array elements (only 0 and 1)\n";
+    int arr[n];
```
Ashborn-SM, line 38: Don't use Variable-length-arrays(VLA) instead use vectors from the STL. Since C++11, VLA is not supported, although there are some compilers that do support VLAs in c++ as an extension(eg: GCC).

Ashborn-SM (commented):
@samughodake please do the changes that are mentioned.

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
samughodake: Hello @Ashborn-SM,
Did all the changes you mentioned

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
```diff
@@ -0,0 +1,62 @@
...
+  int n;
+  cout << "Enter number of array elements\n";
+  cin >> n;
+  vector<int> v(n);
+  cout << "Enter the array elements (only 0 and 1)\n";
+
+  for (int i = 0; i < n; i++)
+  {
+    cin >> v[i];
+    v.push_back(v[i]);
```
Ashborn-SM, line 52: You are actually inserting the inputs twice to the vectors. Since you are using `n` and not `v.size()` in the loop your program works fine because your first `n` values will be the same as the input values you give.
```C
// If this is your inputs then you vector will look like the one below.
n -> 5
inputs: 1 0 1 0 1 

v.size() -> 10
vector -> {1 0 1 0 1 1 0 1 0 1}
```
```suggestion
    cin >> v[i];
```

Ashborn-SM (changes requested):
@samughodake please do the changes mentioned.

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
Ashborn-SM: You should keep in mind that you can only use `cin >> v[i]` if you declare the vector with some size `n`, doing otherwise will cause a [segmentation fault](https://en.wikipedia.org/wiki/Segmentation_fault).

This is because when you declare it with size `n` ,during the compile time the memory of size `n*sizeof(data_type)` will be allocated for the vector and all the memory spaces will be initialised to `0` or a specified value, hence using `cin` or initialising it (eg:`v[1] = 5`) will be valid.

Doing such things without declaring the vector size will not be valid because no memory will be allocated for it to use, so initialising the vector values like `v[i] = i` will not be valid and causes `segfault`

In this case you can make use of `push_back()`. Whenever you try to `push_back()` , the vector will see if the present capacity is enough to accommodate that element, if it's enough the vector will assign the element in the next memory space, if not it will reallocate the memory and copy all the element from the present memory to the newly allocated memory.

Ashborn-SM (approved):
Good Work!!

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
samughodake: Thanks for the above description. I already made changes to the code.

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
```diff
@@ -0,0 +1,61 @@
+/*
```
ming-tsai, line 1: The file name should be `segregate-0-and-1.cpp`

ming-tsai (changes requested):
Hi @samughodake,
Please rename the file name

algorithms/CPlusPlus/Arrays/segregate0and1.cpp
samughodake: Hi @ming-tsai 
Renamed the file as mentioned as well updated the readme file.

ming-tsai (approved):

PR #489, opened 2021-09-26
chore(CSharp): Implemented Dijkstras Algorithm to Find the Shortest Path

Waqar-107:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Added Dijkstras algo with unit tests

**Other information:**

### DIFF
--- a/algorithms/CSharp/src/Graph/dijkstra.cs
+++ b/algorithms/CSharp/src/Graph/dijkstra.cs
@@ -2,7 +2,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace Algorithms.Graph
 {
@@ -54,7 +53,7 @@ namespace Algorithms.Graph
                 {
                     int u = pq.Dequeue();
 
-                    if(visited[u] == true)
+                    if(visited[u] == true || !_edges.ContainsKey(u))
                     {
                         continue;
                     }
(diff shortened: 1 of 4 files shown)


Waqar-107:
@ming-tsai I have used a `Nuget` package here that will cause build failure unless I add the `csproj`. I have added it. Let me know your thoughts on this!

algorithms/CSharp/src/Graph/dijkstra.cs
```diff
@@ -0,0 +1,132 @@
+﻿using Priority_Queue;
```
ming-tsai, line 1: This library is needed? because if found this for [C#](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-6.0&viewFallbackFrom=netstandard-2.1)

Waqar-107: @ming-tsai it seems to me that this is available for `.Net 6.0`. Couldn't use it in `Visual Studio 2019` or [.Net Finddle](https://dotnetfiddle.net/). Any idea how to use it? If I can get a compiler than I could replace the library

ming-tsai: Why do we need this library?

Waqar-107: I couldn't use `PriorityQueue` from `System.Collections.Generic` that is why I used the Priority Queue from this library. If you can help me with using `PriorityQueue` from `System.Collections.Generic` then I can remove this

ming-tsai: > @ming-tsai  I couldn't use `PriorityQueue` from `System.Collections.Generic` that is why I used the Priority Queue from this library. If you can help me with using `PriorityQueue` from `System.Collections.Generic` then I can remove this

Could you tell me why you need `PriorityQueue`?

Waqar-107: > > @ming-tsai  I couldn't use `PriorityQueue` from `System.Collections.Generic` that is why I used the Priority Queue from this library. If you can help me with using `PriorityQueue` from `System.Collections.Generic` then I can remove this
> 
> Could you tell me why you need `PriorityQueue`?

I need a min-heap to implement the algorithm. That is why using a `PriorityQueue`

ming-tsai: Ok, I will check it out

ming-tsai (approved):

PR #508, opened 2021-10-02
chore(CPlusPlus): added search insert position

samughodake:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Arrays/search-insert-position.cpp
+++ b/algorithms/CPlusPlus/Arrays/search-insert-position.cpp
@@ -20,7 +20,7 @@ int searchInsert(vector<int>& nums, int target) {
     int end=nums.size()-1;
     while(start<=end){
         //calculating mid element
-        int mid = (start+end)/2;
+        int mid = start + (end-start)/2;
         //if target found, return the index
         if(nums[mid]==target){
             return mid;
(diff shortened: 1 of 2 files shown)


Capt-Madhan (dismissed):

Ashborn-SM (changes requested):
@samughodake please do the changes mentioned.

samughodake:
Hello, @Ashborn-SM 
I made all the desired changes.
Thanks.

algorithms/CPlusPlus/Arrays/search-insert-position.cpp
```diff
@@ -0,0 +1,69 @@
...
+    cin>>n;
+    cout<<"Enter the array elements\n";
+    vector<int> nums(n);
+    for(int i=0;i<n;i++){
+        cin>>nums[i];
+    }
+    cout<<"Enter the target element\n";
+    int target;
+    cin>>target;
+    cout<<"The target element can be inserted at: "<<searchInsert(nums,target);
```
Ashborn-SM, line 53: The text here can be misleading as it prints `The target element can be inserted at: (some_index)` regardless of whether the target is present or not.

You don't want to print this if the target is present, in such cases a valid statement will be 
`The location of the target is (some_index)`.

My suggestion is to have a valid statement for both the cases(when the target is present & when the target is not present), you can make changes in the function itself accordingly.

Ashborn-SM (approved):
Good Work!!

PR #517, opened 2021-10-04
chore(CPlusPlus): display longest name

samughodake:

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Strings/longest-name.cpp
+++ b/algorithms/CPlusPlus/Strings/longest-name.cpp
@@ -1,0 +1,45 @@
+/*
+Description: Given a list of names, display the longest name.
+
+Approach: To use length() method for every array element
+Keeping the index of maximum length string in max variable
+
+Time Complexity: O(n)
+*/
+
+#include <iostream>
+
+using namespace std;
+//function starts
+string longest(string names[], int n) { //storing the index of max-length string
+  int max = 0;
+
+  for (int i = 1; i < n; i++) {
+    if (names[i].length() > names[max].length()) {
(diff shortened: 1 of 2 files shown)


algorithms/CPlusPlus/Strings/longest-name.cpp
```diff
@@ -0,0 +1,45 @@
...
+  string names[] = {
+    "hi",
+    "hello",
+    "helloall",
+    "helloeveryone"
+  };
+  //calculating size of the array
+  int n = sizeof(names) / sizeof(names[0]);
+  cout << "The longest string in the array is: " << longest(names, n);
+}
```
UG-SEP, line 38: add return 0 to tell that your program runs successfully

UG-SEP (changes requested):
Please make the changes shown below

algorithms/CPlusPlus/Strings/longest-name.cpp
Arsenic-ATG: The body of the `main ()` does not need to contain the return statement, if control reaches the end of main without encountering a return statement, the effect is that of executing `return 0;` 

Although it is recommended to explicitly add it for readability reasons but is not mandatory.

samughodake: Added return 0 to main.
Thanks

UG-SEP (approved):
Great work @samughodake

PR #677, opened 2022-01-20
chore(CPlusPlus): add return string subsequences

vivekthedev:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Return String Subsequences using Recursion

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Recursion/string-subsequences.cpp
+++ b/algorithms/CPlusPlus/Recursion/string-subsequences.cpp
@@ -15,6 +15,7 @@ before by adding single character to the previous call.
 */
 
 #include <iostream>
+#include <string>
 using namespace std;
 int str_subsequences(string input, string output[])
 {


algorithms/CPlusPlus/Recursion/string-subsequences.cpp
```diff
@@ -0,0 +1,45 @@
...
+Appraoch:
+Subsequences are the every possible sequences that can be formed from a string
+from a empty string to every character in the string.
+
+In the below code we return the array size which contains our subsequences.
+In every recursion call the function get the double subsequences of what it was
+before by adding single character to the previous call.
+*/
+
+#include <iostream>
```
Arsenic-ATG, line 17:
```suggestion
#include <iostream>
#include <string>
```

Make sure to include all the required headers, even though some implementations include string header which implementing parts of iostream, but they are not obligated to do so.

algorithms/CPlusPlus/Recursion/string-subsequences.cpp
```diff
@@ -0,0 +1,45 @@
...
+    {
+        output[i + smallOutputSize] = input[0] + output[i];
+    }
+    return 2 * smallOutputSize;
+}
+
+int main()
+{
+    string s = "abc";
+    string *out = new string[1000];
```
Arsenic-ATG, line 38: what if the number of substrings is more than 1000 ?

An alternative here would be to use a flexible container ( like std::vector ) and let it do the heavy lifting of memory management while you take care of this program.

Arsenic-ATG (changes requested):

Arsenic-ATG:
Along with changes suggested, make sure you also update the required README.md index of Cplusplus dir with your creation in it.

ming-tsai:
Hi @vivekthedev,
Any update?

PR #696, opened 2022-01-31
chore(CPlusPlus): add print all nodes at a level on trees

aarushimaurya:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Added a new algorithm in trees- printing all nodes at a level in binary tree

**Other information:**


algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
```diff
@@ -0,0 +1,61 @@
+#include<bits/stdc++.h>
+using namespace std;
+
+typedef struct Node
```
aayushjain7, line 4: please provide a description of the program

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
```diff
@@ -0,0 +1,61 @@
...
+	root->right=new node(3);
+	root->right->left=new node(7);
+	root->right->right=new node(9);
+	root->right->left->left=new node(21);
+	root->right->right->right=new node(11);
+
+	int k = 3;  // here we have taken k=3 (third level of tree)
+	
+	levelorder(root,k);   //calling level order function
+}
```
aayushjain7, line 61: provide a test case with time complexity

aayushjain7 (changes requested):

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
aarushimaurya: I have added a description.

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
aarushimaurya: I have added a test case and time and space complexities.

algorithms/CPlusPlus/README.md
```diff
@@ -137,6 +137,7 @@
 - [Finding the elements of a tree visible from top view](Trees/Top-View-Of-A-Tree.cpp)
 - [Binary Tree Implementation](Trees/binary-tree-implementation.cpp)
 - [Iterative Segment Tree](Trees/IterativeSegmentTree.cpp)
+- [Print all nodes at level k] (Trees/print-all-nodes-at-a-level.cpp)
```
ming-tsai, line 140:
```suggestion
- [Print all nodes at level k](Trees/print-all-nodes-at-a-level.cpp)
```

ming-tsai:
Hi @aayushjain7,
Please check the pull request when you have time.

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
```diff
@@ -0,0 +1,80 @@
...
+	
+	Node(int val)
+	{
+		data= val;
+		left=right=NULL;
+	}
+}node;
+
+//function for level order traversal
+void levelorder(Node* root, int k)
```
Ashborn-SM, line 32: Maybe in C++ this is valid(not really sure) but in C it will throw in an `unkown type name` error, since the name that comes after the brackets is the new typename which is `node` and not `Node` (there exist only `struct Node` not `Node` alone). It would be correct to write `struct Node` here but that would defeat the whole point of `typedef` . It's better to stick with the new typename(`node`) you defined to avoid any confusion.

Also change it everywhere in you code.
```suggestion
void levelorder(node* root, int k)
```

Ashborn-SM (changes requested):

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
```diff
@@ -0,0 +1,80 @@
...
+	int data;
+	struct Node* left;
+	struct Node* right;
+	
+	Node(int val)
+	{
+		data= val;
+		left=right=NULL;
+	}
+}node;
```
Ashborn-SM, line 29: In c++ there's no difference between `struct` and `typedef struct` , since unlike C in C++ the `struct` keyword is optional which makes it effectively same as `typedef struct` in C.
```suggestion
struct Node
{
	int data;
	struct Node* left;
	struct Node* right;
	
	Node(int val)
	{
		data= val;
		left=right=NULL;
};
```

Ashborn-SM (changes requested):

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
```diff
@@ -0,0 +1,80 @@
...
+		for(int i=0; i<n; i++)
+		{
+			Node* temp=q.front();
+			q.pop();
+			if(count==k)  //if level is equal to the required level, then print its nodes
+				cout<<temp->data<<" ";
+			if(temp->left!=NULL)
+				q.push(temp->left);
+			if(temp->right!=NULL)
+				q.push(temp->right);
```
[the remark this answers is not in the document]
aarushimaurya: You're right @Ashborn-SM

algorithms/CPlusPlus/Trees/print-all-nodes-at-a-level.cpp
aarushimaurya: Won't I have to write struct Node everywhere instead of just node then?

Ashborn-SM (dismissed):

PR #699, opened 2022-02-04
chore(Python): add Rabin Karp algorithm

kartiknad:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Added a new algorithm, Rabin-Karp pattern matching algorithm implemented in python

**Other information:**

### DIFF
--- a/algorithms/Python/strings/rabin-karp-algorithm.py
+++ b/algorithms/Python/strings/rabin-karp-algorithm.py
@@ -2,8 +2,8 @@
 String pattern matching algorithm which performs efficiently for large text and patterns
 
 Algorithm: 
-Rabin Karp works on the concept of hashing. The if the substring of the given text 
-is same as the pattern then the corresponding hash value should also be same. Exploting this idea 
+Rabin Karp works on the concept of hashing. If the substring of the given text 
+is same as the pattern then the corresponding hash value should also be same. Exploiting this idea 
 and designing a hash function which can be computed in O(m) time for both pattern and initial window 
 of text. The subsequent  window each will require only O(1) time. And we slide the window n-m times 
 after the initial window. Therefore the overall complexity of calculating hash funtion for text is O(n-m+1)
@@ -29,7 +29,7 @@ def rabin_karp(T: str, P: str, q: int ,d: int = 256) -> None :
 
             q: int
                An appropriately chosen prime number based on length of input strings
-               The higher the prime number, the lower the collisiona nd spurious hits
+               The higher the prime number, the lower the collisions and spurious hits
                
             d: int, default value 256
                Denotes the no of unique character that is used for encoding
@@ -87,7 +87,7 @@ def rabin_karp(T: str, P: str, q: int ,d: int = 256) -> None :
         print("Pattern not found")
     else:
         for i in pos:
-            print("Pattern found at pos: {}".format(i))
+            print(f"Pattern found at pos: {i}")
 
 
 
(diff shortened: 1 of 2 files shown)


algorithms/Python/strings/rabin-karp-algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+    
+    if not found: # If pattern not found in text
+        pos.append(-1)
+        
+    #Printing results
+    if pos[0]==-1:
+        print("Pattern not found")
+    else:
+        for i in pos:
+            print("Pattern found at pos: {}".format(i))
```
Arsenic-ATG, line 90: The pyupgrade test is most probably failing due to use use of old style string formatting here, and recommends to use f-strings instead 

```suggestion
            print(f"Pattern found at pos: {i}")
```

algorithms/Python/strings/rabin-karp-algorithm.py
```diff
@@ -0,0 +1,101 @@
+'''
+String pattern matching algorithm which performs efficiently for large text and patterns
+
+Algorithm: 
+Rabin Karp works on the concept of hashing. The if the substring of the given text 
+is same as the pattern then the corresponding hash value should also be same. Exploting this idea 
+and designing a hash function which can be computed in O(m) time for both pattern and initial window 
+of text. The subsequent  window each will require only O(1) time. And we slide the window n-m times 
+after the initial window. Therefore the overall complexity of calculating hash funtion for text is O(n-m+1)
+Once the hash value matches, the underlying string is again checked with pattern for matching
```
Arsenic-ATG, line 10:
```suggestion
Algorithm: 
Rabin Karp works on the concept of hashing. If the substring of the given text 
is same as the pattern then the corresponding hash value should also be same. Exploiting this idea 
and designing a hash function which can be computed in O(m) time for both pattern and initial window 
of text. The subsequent  window each will require only O(1) time. And we slide the window n-m times 
after the initial window. Therefore the overall complexity of calculating hash funtion for text is O(n-m+1)
Once the hash value matches, the underlying string is again checked with pattern for matching
```

algorithms/Python/strings/rabin-karp-algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+    
+            T: string
+               The string where the pattern needs to be searched
+
+            P: string 
+               The pattern to be searched
+
+            q: int
+               An appropriately chosen prime number based on length of input strings
+               The higher the prime number, the lower the collisiona nd spurious hits
```
Arsenic-ATG, line 32:
```suggestion
               The higher the prime number, the lower the collision and spurious hits
```

Arsenic-ATG (changes requested):
The suggested changes should make the PR pass all the pre-tests.

ming-tsai:
Hi @kartiknad,
Any update?

kartiknad:
> Hi @kartiknad, Any update?

The suggested changes have been made

algorithms/Python/strings/rabin-karp-algorithm.py
kartiknad: Thanks for the suggestions, I have made the changes

Arsenic-ATG (approved):
My last commit involve fixing some more nit-picks detected by pyupgrade and codespell checks. 

the PR should be good to go now.

PR #718, opened 2022-03-19
chore(C): add radix sort

mohanadtalat91:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Add radix sort algorithm in C language  
[The online resource](https://en.wikipedia.org/wiki/Radix_sort)
**Other information:**


algorithms/C/sorting/radix-sort.c
```diff
@@ -0,0 +1,67 @@
...
+    for(int i=0; i<n; i++){
+        if(arr[i] > myMax)myMax = arr[i] ;
+    }
+
+    return myMax ;
+}
+
+void count_sort(int arr[], int n, int digit){
+    int count[10] = {0} ;
+    int sorted[n] ;
```
Arsenic-ATG, line 16: I would not recommend using a Variable Length Array (VLA) here.
Although C99 and above C standards accept it, it is still considered as a code smell and can cause troubles when used with large value of `n`.

A heap allocated array should do the trick here.

algorithms/C/sorting/radix-sort.c
```diff
@@ -0,0 +1,67 @@
...
+        //          Yes      :     Yes      :       No 
+
+    int arr[] = {374, 42, 6, 920,825} ;
+    int n = sizeof(arr)/sizeof(int) ;
+
+    printf("Before sorting : ") ;
+    for(int i=0; i<n; i++){
+        printf("%d",arr[i] );
+        printf(" ") ;
+    }
```
Arsenic-ATG, line 57: this can simply be boiled down to reduce clutter 
```suggestion
    for(int i=0; i<n; i++)
        printf("%d ",arr[i] );
```

algorithms/C/sorting/radix-sort.c
```diff
@@ -0,0 +1,67 @@
...
+
+    radix_sort(arr, n) ;
+
+    printf("\n\nAfter sorting : ") ;
+    for(int i=0; i<n; i++){
+        printf("%d",arr[i] );
+        printf(" ") ;
+    }
```
Arsenic-ATG, line 65: same here
```suggestion
    for(int i=0; i<n; i++)
        printf("%d ",arr[i] );
```

Arsenic-ATG (changes requested):
Apart from the changes mentioned in review comments, every addition of new algorithm should also correspond to updating the required index README.md ( C in this case ).

algorithms/C/sorting/radix-sort.c
mohanadtalat91: But all resources use the same idea that I'm used above !? 
Anyway could you please tell me what'd U recommend `calloc()` or `malloc()` ?

Arsenic-ATG:
Nice, now all that is left is to updating the C Readme.md file and the PR should be good to go.

mohanadtalat91:
> Nice, now all that is left is to updating the C Readme.md file and the PR should be good to go.

Updated. 
@Arsenic-ATG

Arsenic-ATG (approved):

PR #747, opened 2022-04-23
Add Hash Table data structures

mohanadtalat91:
**PR Checklist:**
- [x] My submission is formatted according to the guidelines in the [contributing guide](https://github.com/MakeContributions/DSA#contribution-guidelines)
- [x] My addition is on refer on the language README.md file
- [x] My addition does not have a spelling problem
- [x] My submission has a proper and user-friendly description of the algorithm
- [x] My submission has the time complexity of the algorithm
- [x] My submission has sample input-output of the program (NOT FOR PYTHON)

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] New algorithm
- [ ] Optimization in previous algorithms
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Briefly describe the changes in this PR**

Implementation of Hash Table in Cpp with different hashing functions, Implement two ways to solve colliosion

**Other information:**

- Implementation of Hash Table with different hashing functions, Implement two ways to solve colliosion  : 

	 - [x] Chaining 
	 - [x] Linear Probing

PR #754, opened 2022-04-28
Add sparse array

mohanadtalat91:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

Add sparse array using linked list data structure. 

**Other information:**

If most of the elements of the array have 0 value, then it is called a sparse array. Why to use Sparse array instead of simple array, because Storage: There are lesser non-zero elements than zeros and thus lesser memory can be used to store only those elements.


Arsenic-ATG (changes requested):
changes suggested in #747 also apply here

PR #795, opened 2022-08-08
Create NKnights.java

Faizan711:
This is a Java Program to find all the possible combinations to place N Knights in an N X N Chessboard such that they all are non-attacking to each other.

Please check my program.

### DIFF
--- a/algorithms/Java/backtracking/NKnights.java
+++ b/algorithms/Java/backtracking/NKnights.java
@@ -1,0 +1,75 @@
+public class NKnights {
+    public static void main(String[] args) {
+        int n=3;
+        boolean[][] board=new boolean[n][n];
+        System.out.println(2*knight(board,0,0,3));
+    }
+    static int knight(boolean[][] board,int row,int col,int tar){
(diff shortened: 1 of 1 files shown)


Ashad001 (changes requested):
Knight can move on 8 places on a board at a time, you have only checked for 4. The other four might be:
* row + 2 , col + 1
* row + 1 , col + 2
* row + 2 , col - 1
* row + 1 , col - 2
Correct me if I am wrong.

Faizan711:
This solution is for placing "N" Knights only in a" N*N "chessboard.

Ashad001:
Have a look on this picture
image
In your program, you have checked for 4 moves but actually you have to check for all 8 possible positions.

Ashad001:
Sorry, I got it, Since you are backtracking (recursion). We actually don't have to check for lower positions: we are placing knights row by row. So checking for each row we only have to check for upper positions.
Suggestion: Can you add description of your algorithm in the code. that would be helpful. THANKS

Issue #808, opened 2022-08-19
Binary Power in C++


mohnishdeshpande:
## Information about Algorithm

Faster computing of powers (a^b) using Binary Exponentiation.
### Have you read the [Contributing.md](https://github.com/MakeContributions/DSA/blob/main/CONTRIBUTING.md) and [Code of conduct](https://github.com/MakeContributions/DSA/blob/main/CODE_OF_CONDUCT.md)

- [x] Yes
- [ ] No

SubhamPanda2003:
Hey,I am willing to work on this, please assign it to me.

ming-tsai:
Hi @mohnishdeshpande, @angshudas and @Arsenic-ATG,
We have both pull requests #810 and #815 working on this, please make sure which should be closed and make pair programming in another pull request.

mohnishdeshpande:
Hi @ming-tsai,
In #815, there are a fixes changes required.
On line 16:
-long ans= power(base , expo/2);
+long ans= power_recursion(base , expo/2);
and line 33:
-res = res * expo;
+res = res * base;

To solve the issue here, I have added the recursive implementation from @angshudas to #810 , sourcing his account in comments.

You can review and merge #810 if you are happy with it.
P.S. I have tested it already.

Arsenic-ATG:
> Hi @mohnishdeshpande, @angshudas and @Arsenic-ATG,
> We have both pull requests #810 and #815 working on this, please make sure which should be closed and make pair programming in another pull request.

According to me, the one opened earlier should be given preference (i.e.  #810 ). 

@angshudas , if you want to make changes, then a preferable way might be to work on source branch of previous pull request ( [mohnishdeshpande:808](https://github.com/mohnishdeshpande/DSA/tree/808) in this case ). It will keep the changes regarding one algorithm isolated to one PR only (at least on the main repository).

Or 

Update the existing file once the previous one is merged (as refactoring/optimisation)

Arsenic-ATG:
> To solve the issue here, I have added the recursive implementation from @angshudas to #810 , sourcing his account in comments.

I guess a much better alternative would be to mark @angshudas as a co-author of the commit adding recursive version of the algorithm, like this, version control will also be able to keep track of original author of the changes in the repo ( would be helpful if someone does a `git blame` on the file )

PR #810, opened 2022-08-20
Add binary-power.cpp to CPlusPlus/Maths (#808)

mohnishdeshpande:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Maths/binary-power.cpp
+++ b/algorithms/CPlusPlus/Maths/binary-power.cpp
@@ -1,0 +1,39 @@
+#include<iostream>
+using namespace std;
+
+/*
+The idea is to half the power on every iteration
+Computes the power in log(n) operations
+
+On every iteration:
+Square the base, half the power
+
+Special case - if power is odd:
+Multiply the ans with a
+Because the ODD-NUMBER % 2 == 1
+
+Note: There will always be one iteration where power is odd
+*/
+
+long binpow(long a, long b){
+	long ans=1;
+	// while b is greater than zero, we continue the binary exponentiation
+	while(b>0){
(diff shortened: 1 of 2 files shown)


Ashad001 (commented):
Great!

algorithms/CPlusPlus/Maths/binary-power.cpp
```diff
@@ -0,0 +1,39 @@
...
+Square the base, half the power
+
+Special case - if power is odd:
+Multiply the ans with a
+Because the ODD-NUMBER % 2 == 1
+
+Note: There will always be one iteration where power is odd
+*/
+
+long binpow(long a, long b){
```
Arsenic-ATG, line 18: What if the result is bigger than the value that `long` can hold ?

mohnishdeshpande: Unfortunately 64 bit Integer is the limit in standard C++. However, if you are dealing with positive numbers you can replace `long` with `unsigned long` to double the limit.z

Arsenic-ATG: > Unfortunately 64 bit Integer is the limit in standard C++. However, if you are dealing with positive numbers you can replace `long` with `unsigned long` to double the limit.z

I think changing to `unsigned long` would make the result restrict (wrap) to a much more meaningful numeric field ( an alternative would be to modulus to a prime number )

Arsenic-ATG (approved):

PR #818, opened 2022-08-25
chore(CPlusPlus): add sparse matrix

ssandra102:
**PR Checklist:**

- [ ] Bugfix
- [x] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/CPlusPlus/Arrays/sparse_matrix.cpp
+++ b/algorithms/CPlusPlus/Arrays/sparse_matrix.cpp
@@ -1,40 +1,96 @@
 /*Description :c++ solution to check if a given matrix is sparse matrix.
 If the count of zeroes present in the mmatrix is more than half the elements of the matrix, 
-it is flagged as a sparse matrix.*/
+it is flagged as a sparse matrix.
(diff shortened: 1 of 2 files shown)


mohitchakraverty (commented):
This looks fine which only checks for sparse matrix or not but also if you can add its triple form representation because representing a sparse matrix by a 2D array leads to wastage of lots of memory as zeroes in the matrix are of no use in most of the cases. So, instead of storing zeroes with non-zero elements, we create triples- (Row, Column, value) in which we only store non-zero elements. And instead, use this for different purposes.

zaksandra:
@mohitchakraverty thank you for reviewing, I'll add the triple form representation as well.

Arsenic-ATG (changes requested):
Please also update the proper README.md with your program in the index. 


for your case, it would be [algorithms/CPlusPlus/README.md](https://github.com/ssandra102/DSA/blob/add-new-code/algorithms/CPlusPlus/README.md).

Arsenic-ATG (approved):

Arsenic-ATG (approved):

PR #843, opened 2022-09-15
Update shuffle_array.c

LUCASFRANKINC:
Since you are using the array pointer, just increment it in the loop and access the value in the address.

**PR Checklist:**
- [ ] My submission is formatted according to the guidelines in the [contributing guide](https://github.com/MakeContributions/DSA#contribution-guidelines)
- [ ] My addition is on refer on the language README.md file
- [ ] My addition does not have a spelling problem
- [ ] My submission has a proper and user-friendly description of the algorithm
- [ ] My submission has sample input-output of the program (NOT FOR PYTHON)

- [ ] Bugfix
- [ ] New algorithm
 [ ] Optimization in previous algorithms
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**

### DIFF
--- a/algorithms/C/arrays/shuffle_array.c
+++ b/algorithms/C/arrays/shuffle_array.c
@@ -33,8 +33,9 @@ void shuffle(int *array, int size) // Randomly shuffles given array
 
 void print_array(int *array, int size) // Printing array
 {
-    for (int i = 0; i < size; i++) {
-        printf("%d ", array[i]);
+    //PERF: Since you are using the array pointer, just increment it in the loop.
+    for (int i = 0; i < size; i++, array++) {
+        printf("%d ", *array);
     }
     putchar('\n');
 }


algorithms/C/arrays/shuffle_array.c
```diff
@@ -33,8 +33,9 @@ void shuffle(int *array, int size) // Randomly shuffles given array
 
 void print_array(int *array, int size) // Printing array
 {
-    for (int i = 0; i < size; i++) {
-        printf("%d ", array[i]);
+    //PERF: Since you are using the array pointer, just increment it in the loop.
+    for (int i = 0; i < size; i++, array++) {
+        printf("%d ", *array);
```
LUCASFRANKINC, line 38: Increment the array address, then dereference the pointer to output the value in the address.

LUCASFRANKINC (commented):
Rather than having to use the bracket access whereas we called the array pointer, let the work be handled by the pointer and leave the syntax as clean as possible. Just increment the address for they are contiguous, access the value via de-referencing the pointer.

Ashad001 (changes requested):
Hi @LUCASFRANKINC, However you have updated the code please make sure to use the same style for all the occurrences as it will look more cleaner i.e. in function `shuffle`
Please update this and comment the previous one so the people may see the difference as well. Thank you!

PR #976, opened 2022-10-07
chore(C): added counting sort program

anika-kamath:
**PR Checklist:**
- [ ] My submission is formatted according to the guidelines in the [contributing guide](https://github.com/MakeContributions/DSA#contribution-guidelines)
- [ ] My addition is on refer on the language README.md file
- [ ] My addition does not have a spelling problem
- [ ] My submission has a proper and user-friendly description of the algorithm
- [ ] My submission has sample input-output of the program (NOT FOR PYTHON)

- [ ] Bugfix
- [ ] New algorithm
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Other, please describe:

**Other information:**


algorithms/C/sorting/counting-sort.c
```diff
@@ -0,0 +1,33 @@
+#include<stdio.h>
+ 
+void counting_sort(int a[],int n,int max)
+{
+     int count[50]={0},i,j;
+    
+     for(i=0;i<n;++i)
+      count[a[i]]=count[a[i]]+1;
```
Arsenic-ATG, line 8: What happens when the value of `a[i]` is larger than 50 ?

algorithms/C/sorting/counting-sort.c
```diff
@@ -0,0 +1,33 @@
...
+    
+     for(i=0;i<=max;++i)
+      for(j=1;j<=count[i];++j)
+       printf("%d ",i);
+}
+ 
+int main()
+{
+    int a[50],n,i,max=0;
+    printf("Enter number of elements: ");
```
Arsenic-ATG, line 20: If the buffer length ( `n` in this case ) is meant to be taken interactively, then why initialise it with a fixed size before ? Wouldn't it cause problems involving insufficient space if the value of `n` is larger than the 50 or space wastage if it's less than that.

It's recommended to

algorithms/C/sorting/counting-sort.c
```diff
@@ -0,0 +1,33 @@
+#include<stdio.h>
+ 
+void counting_sort(int a[],int n,int max)
```
Arsenic-ATG, line 3: If the function doesn't alter the values of `n` and `max` then wouldn't it be better for program to explicitly mark them as constants ? It would not only help in making this function callable with const values but would also avoid any accidental changes in future.

```suggestion
void counting_sort(int a[], const int n, const int max)
```

Arsenic-ATG (changes requested):

PR #1048, opened 2022-10-19
chore(CPlusPlus): add special index in an array

nandinisahu407:
i have contributed by how to find special index in an array 
i.e if an element is deleted from index I  then sum of even index=sum of odd index
(it contains the best approach to solve this problem ,looking forward to contribute more)

### DIFF
--- a/algorithms/CPlusPlus/Arrays/specialindex2.cpp
+++ b/algorithms/CPlusPlus/Arrays/specialindex2.cpp
@@ -14,7 +14,8 @@ int main(){
     int num;
     cout<<"enter length"<<endl;
     cin>>num;
-    int arr[num];
+    vector <int> arr (num);
+
     for(int i=0;i<num;i++){
         cin>>arr[i];
     }
--- a/algorithms/CPlusPlus/Arrays/specialindex2.cpp
+++ b/algorithms/CPlusPlus/Arrays/specialindex2.cpp
@@ -1,0 +1,55 @@
+/*
+@author: nandinisahu407
+special index-> if after deleting element from index i , sum of even index=sum of odd index
+approach->
+after deleting ,previous element at odd index will be now at even index and vice versa
+s_odd= odd[0 to i]+ even[i+1 to len]
+s_even=even[0 to i]+odd[i+1 to len]
+
+*/
+#include<iostream>
+using namespace std;
+ 
+int main(){
+    int num;
+    cout<<"enter length"<<endl;
+    cin>>num;
+    int arr[num];
+    for(int i=0;i<num;i++){
+        cin>>arr[i];
+    }
+
+    int count=0;
+    int s_even,s_odd;
+    for(int i=0;i<num;i++){ //checking whether special index or not
+        s_even=0,s_odd=0;
+
+        for(int j=0;j<i;j++){   // from index[0,i]
+            if(j%2==0){
+                s_even+=arr[j];
+            }
+            else{
+                s_odd+=arr[j];
+            }
+        }
+        for(int j=i+1;j<num;j++){    //from index[i+1,len]
+            if(j%2==0){
+                s_odd+=arr[j];
+            }
+            else{
+                s_even+=arr[j];
+            }
+        }
+
+        if(s_even==s_odd){    //checking whether sum of even index=sum of odd index
+            cout<<"\n special index found at: "<<i;
+            count++;
+        }
+        else{
+            continue;
+        }
+
+    }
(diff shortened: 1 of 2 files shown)


algorithms/CPlusPlus/Arrays/specialindex2.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+*/
+#include<iostream>
+using namespace std;
+ 
+int main(){
+    int num;
+    cout<<"enter length"<<endl;
+    cin>>num;
+    int arr[num];
```
Arsenic-ATG, line 17: Why not use a much more safe and reliable stl container instead of using a VLA ( variable length array ) here and relying on the particular C++ implementation to support it as a non-standard extension ?

algorithms/CPlusPlus/Arrays/specialindex2.cpp
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+        for(int j=i+1;j<num;j++){    //from index[i+1,len]
+            if(j%2==0){
+                s_odd+=arr[j];
+            }
+            else{
+                s_even+=arr[j];
+            }
+        }
```
Arsenic-ATG, line 42: just a minor suggestion: wouldn't it be better to use one single loop with an additional check in it to compute the required sums ? 
It would not only reduce the overall code footprint but would also remove redundancy in the program.

```suggestion
        for (int j = 0; j < num; j++) {
            if (j != i) {
                if (j % 2 == 0) {
                    s_odd += arr[j];
                } 
                else {
                    s_even += arr[j];
                }
            }
        }
```

Arsenic-ATG (changes requested):
apart form the changes suggested in the review comments, it's also mandatory for every new-algorithm to have a corresponding entry in the appropriate index README.md ( check previously merged pull requests for examples ).

algorithms/CPlusPlus/Arrays/specialindex2.cpp
nandinisahu407: @Arsenic-ATG the suggested code is not giving the correct output
> my program is to find special index in an array . special index-> when element is deleted from index (eg i ) then to be a special index the sum of even indexes =sum of odd index **after deletion**
 eg->   index :    0     1     2      3      4       5
             arr:         4     3     2     7      6       -2

            to check  whether index 2 is a special index or not
           after deleting element from index 2

             index:     0      1      2       3      4 
             arr:         4      3       7      6       -2

            sum_even=4+7+-2=9      (adding element present in even indexes)
           sum_odd=3+6=9
          sum_even=sum_odd        hence index 2-> is special index

APPROACH-> we could have shifted every element after index i and could have directly add even and odd indexes
but in c++, shifting elements is a very tiring job .so i have used a very smart approach to solve this problem without shifting elements in array.
we can notice **after deleting element** from index 2-> 

- elements **before index 2** remained unchanged
            s_even+=element at even index
            s_odd+=element at odd index

- but elements **after index  2** shifted by 1 to their left (but we have not shifted elements in arr,we are looking for pattern)
 i.e-> elements  before at odd place is now at even place (eg 7,-2)
                              s_even+=element at odd index in original array

         elements before at  even  place is now at odd place(eg. 6)
                            s_odd+=element at even index in original array
hence ,i have used two loops .one for element before index 2 and other for elements after index 2.
(p.s-> index 2 is just for example but similarly we have to check for every index in array)

looking forward to more reviews in this program to improvise it more ,please let me know how we can do it in a better way than this approach in array.
(p.s i don't know how to use vectors yet but will learn soon and try with it also)

Arsenic-ATG: hmm, make sense
Thanks for pointing that out.

as for vectors are concerned, it's extremely simple to use and contains appropriate overloads to access it like an array ( via `[]` operator ), so in your program it should be as simple as just simple replacing the declaration of the array ( on line 17 ) to a vector and almost nothing would require changing in the entire program ( let me suggest some changes there ).

I fear alternative approach to a vector container would be to dynamically allocating and managing the array at runtime.

Arsenic-ATG: FWIF: you can check out [vector on cppreference](https://en.cppreference.com/w/cpp/container/vector) to know more about what std::vector can do.

Arsenic-ATG (commented):
nice, now all that is left is to update the required index READMD.md file and PR should be good to go.

Arsenic-ATG (approved):

Issue #1058, opened 2022-10-20
Bad memory usage in  DSA/algorithms/CPlusPlus/Graphs/dijkstra.cpp 


Denisthatorf:
The Dijkstra algorithm uses vector<vector<pair<int,int>>> graph(V+1,vector<pair<int,int>>()) as a matrix. This is very, very bad for memory, because the vector in the vector in memory is not stored linearly, and can be scattered in memory very much. As an alternative, I propose to use a regular array (data = new T [rows * columns]) allocated on the heap and get its elements like this : *(_data + row * _rows + column).

ming-tsai:
Hi @Denisthatorf,
Could you help me to this bug?

Denisthatorf:
Yes, sure

Denisthatorf:
https://github.com/MakeContributions/DSA/pull/1061 - my pull request for fixing a problem

PR #1061, opened 2022-10-21
enh(CPlusPlus): memory usage on Dijksta algorithm

Denisthatorf:
https://github.com/MakeContributions/DSA/issues/1058 - try to fix this problem
I optimize the Dijkstra algorithm in C++ 

The Dijkstra algorithm uses vector<vector<pair<int,int>>> graph(V+1,vector<pair<int,int>>()) as a matrix. This is very, very bad for memory, because the vector in the vector in memory is not stored linearly, and can be scattered in memory very much. As an alternative, I propose to use a regular array (data = new T [rows * columns]) allocated on the heap and get its elements like this : *(_data + row * _rows + column).

### DIFF
--- a/algorithms/CPlusPlus/Graphs/dijkstra.cpp
+++ b/algorithms/CPlusPlus/Graphs/dijkstra.cpp
@@ -82,7 +82,7 @@ void getExampleMatrix(int*& matrix_out, size_t& size_out);
 
 int main()
 {
-    //I highly recomend to create matrix class
+    //I highly recommend to create matrix class
 
     size_t size;
     std::cout << "Graph Size: ";
--- a/algorithms/CPlusPlus/Graphs/dijkstra.cpp
+++ b/algorithms/CPlusPlus/Graphs/dijkstra.cpp
@@ -1,99 +1,177 @@
-//Dijkstra's algorithm
-//implemented in the context of a directed graph
-#include <bits/stdc++.h>
-using namespace std;
+#include <cstddef>
+#include <limits>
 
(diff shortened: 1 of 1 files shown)


algorithms/CPlusPlus/Graphs/dijkstra.cpp
```diff
@@ -1,99 +1,177 @@
...
-        //otherwise we update our current cost(distance)
-        dist[cVertex]=cCost;
-
-        if(cVertex==end){
-            ret=cCost;
-            break;
+#include <cstddef>
+#include <limits>
+
+//I highly recomend to create matrix class
```
Arsenic-ATG, line 4:
```suggestion
//I highly recommend to create matrix class
```

algorithms/CPlusPlus/Graphs/dijkstra.cpp
```diff
@@ -1,99 +1,177 @@
...
+#include <iostream>
+
+
+//function declaration below
+//Generate and std::cout << matrix
+void getExampleMatrix(int*& matrix_out, size_t& size_out);
+
+int main()
+{
+    //I highly recomend to create matrix class
```
Arsenic-ATG, line 85:
```suggestion
    //I highly recommend to create matrix class
```

Arsenic-ATG (changes requested):
typos causing codespell check to fail 

```
./algorithms/CPlusPlus/Graphs/dijkstra.cpp:4: recomend ==> recommend
./algorithms/CPlusPlus/Graphs/dijkstra.cpp:85: recomend ==> recommend
```

Denisthatorf:
Can you check one more time?

